PR #55, opened 2019-01-18
Split deploy-environment command into create-package and deploy-package

abbotware:
new command **create-package**

updated command  **deploy-environment** 
- now has a package parameter

*Issue #2 *

*Description of changes:*
It's been a year since I opened this issue and have not seen any progress made.  It is extremely easy in principle..  I have **create-package** working more or less, but need some pointers on the **deploy-package** as this will require a new parameter.


src/Amazon.ElasticBeanstalk.Tools/Commands/CreatePackageCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+            var zipArchivePath = Path.Combine(Directory.GetParent(publishLocation).FullName, new DirectoryInfo(projectLocation).Name + "-" + DateTime.Now.Ticks + ".zip");
+
+            this.Logger?.WriteLine("Zipping up publish folder");
+            Utilities.ZipDirectory(this.Logger, publishLocation, zipArchivePath);
+            this.Logger?.WriteLine("Zip archive created: " + zipArchivePath);
+
+            return Task.FromResult(true);
+        }
+
+        private void SetupAWSDeploymentManifest(string publishLocation)
```
normj: This is a duplicate of what is in DeployEnvironmentCommand. Create an EBUtilities class in the parent namespace and move this code to that class. That would make it consistent with the Lambda commands.

src/Amazon.ElasticBeanstalk.Tools/Commands/CreatePackageCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+using ThirdParty.Json.LitJson;
+
+namespace Amazon.ElasticBeanstalk.Tools.Commands
+{
+    public class CreatePackageCommand : EBBaseCommand
+    {
+        public const string COMMAND_NAME = "create-package";
+        public const string COMMAND_DESCRIPTION = "Package the application to zip file to be deployed later";
+
+        public static readonly IList<CommandOption> CommandOptions = BuildLineOptions(new List<CommandOption>
```
normj: Filter this list to the switches that are actually used.

src/Amazon.ElasticBeanstalk.Tools/Commands/DeployPackageCommand.cs
```diff
@@ -0,0 +1,600 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using ThirdParty.Json.LitJson;
+
+namespace Amazon.ElasticBeanstalk.Tools.Commands
+{
+    public class DeployPackageCommand : EBBaseCommand
```
normj: Instead of creating a separate deploy command can we reuse the existing deploy-environment that has an optional package command? If that is set deploy-environment would skip the packaging park.

abbotware:
just an observation, in noticed the used of backing fields

```
        DeployEnvironmentProperties _deployAppProperties;
        public DeployEnvironmentProperties DeployEnvironmentOptions
        {
            get
            {
                if (this._deployAppProperties == null)
                {
                    this._deployAppProperties = new DeployEnvironmentProperties();
                }

                return this._deployAppProperties;
            }
            set { this._deployAppProperties = value; }
        }

```


isn't this more succinct (and thread-safe):

        public DeployEnvironmentProperties DeployEnvironmentOptions  { get; } = new DeployEnvironmentProperties();

src/Amazon.ElasticBeanstalk.Tools/Commands/CreatePackageCommand.cs
abbotware: done in the last commit - the function is not static, so I had to send a few parameters in to make it static

src/Amazon.ElasticBeanstalk.Tools/Commands/CreatePackageCommand.cs
abbotware: done in the last commit

src/Amazon.ElasticBeanstalk.Tools/Commands/DeployPackageCommand.cs
abbotware: might have gotten this working

abbotware:
I can confirm this works perfectly for me

src/Amazon.ElasticBeanstalk.Tools/Commands/CreatePackageCommand.cs
```diff
@@ -0,0 +1,107 @@
...
+namespace Amazon.ElasticBeanstalk.Tools.Commands
+{
+    public class CreatePackageCommand : EBBaseCommand
+    {
+        public const string COMMAND_NAME = "create-package";
+        public const string COMMAND_DESCRIPTION = "Package the application to a zip file to be deployed later";
+
+        public static readonly IList<CommandOption> CommandOptions = BuildLineOptions(new List<CommandOption>
+        {
+            CommonDefinedCommandOptions.ARGUMENT_PROJECT_LOCATION,
```
normj: Remove ARGUMENT_PROJECT_LOCATION because it comes in from the BaseCommand.

src/Amazon.ElasticBeanstalk.Tools/Commands/DeployEnvironmentCommand.cs
```diff
@@ -79,68 +68,70 @@ protected override void ParseCommandArguments(CommandOptions values)
...
-                }
-                catch(Exception e)
-                {
-                    throw new ElasticBeanstalkExceptions("Error creating Elastic Beanstalk application: " + e.Message, ElasticBeanstalkExceptions.EBCode.FailedCreateApplication);
+                        targetFramework = this.GetStringValueOrDefault(this.DeployEnvironmentOptions.TargetFramework, CommonDefinedCommandOptions.ARGUMENT_FRAMEWORK, true);
+                    }
                 }
-            }
 
+                await CreateEBApplicationIfNotExist(application, doesApplicationExist);
```
normj: This line is repeated if the package does exist and if doesn't. Move this line to after the if/else and remove the duplicate from the else.

abbotware:
I need both changesets - I assume the env variable changes are smaller to absorb and assume that will get accepted before this PR (this changes more code)

abbotware:
can you suggest some tests cases?

normj:
This has been released as part of version 3.1.0 of Amazon.ElasticBeanstalk.Tools. Thanks for the PR and patience to get this out.

normj:
Forgot to mention I did change a few things in the package command to make it look similar to the Lambda package command.

* Change the command to be `dotnet eb package` instead of `dotnet eb create-package`
* For the package command you would set the `--output-package` switch to specify the target zip file. When deploying you set the `--package` switch.

PR #57, opened 2019-01-20
Add environment variable replacement to GetStringValueOrDefault

abbotware:
*Description of changes:*
Allows a config file to use $(VARIABLE) that will be replaced with environment variables at run-time.

### DIFF
--- a/src/Amazon.Common.DotNetCli.Tools/Utilities.cs
+++ b/src/Amazon.Common.DotNetCli.Tools/Utilities.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
+using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text;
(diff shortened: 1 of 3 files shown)


abbotware:
I only use config files.. but having to put 'relative pathing'  (..\..\ notation)  everywhere is painful.  I would much rather just use a token that gets expanded at runtime

abbotware:
I can confirm this works perfectly for me.. as long as the {varaible} token is not typically used in any existing parameter/config this should be fine for everyone else too.   If somehow a token appeared, it would get replaced with an empty string as the environment variable won't exist

abbotware:
what the token is - doesnt matter - its a regex.   I just need any form of token replacement so I don't have to make convoluted config files.  I'll change to $(VARIABLE) and make some unit tests

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -13,11 +13,37 @@
...
+        public static string ReplaceEnvironmentVariables(string original)
+        {
+            MatchCollection matches = EnvironmentVariableTokens.Matches(original);
+
+            var modified = original;
+
+            foreach (Match m in matches)
+            {
+                var withoutBrackets = m.Value.Substring(2, m.Value.Length - 3);
+                var env = Environment.GetEnvironmentVariable(withoutBrackets);
```
normj: I'm thinking if the environment variable is not set and env is either null or "" then the code should not do the replace call. That way if somebody did have a legitimate reason for having a $(somestring) that was not an environment variable it would unlikely be removed.

There should also be a test case for the environment variable not being found.

abbotware: Well, there already was a test for this - but one that verifies that it removed the variable - I know from using a lot of MSBuild and other systems, they always replace $(Identifier) with null / empty string.

**However**, its not a big deal to be cautious especially introducing potentially breaking behavior.   In doing so we now have to get ALL variables and test for the absence of the tokens.  We can no longer just call `Environment.GetEnvironment` since the intended value of the variable could be set to empty.string or null...   

Much to my surprise `System.Environment.GetEnvironmentVariables()`  is seems like a .Net Framework 1.0 API as its not even using a Dictionary<string,string>.  At first wanted to cache all variables using a simple static- but the unit tests fail since the static is initialized only once per app domain / test run.

I would say display a warning for un-replaced variables, but I'll leave that up to you.  Line 48 (the continue statement) of Utilties.cs is where you could add the log statement.

normj:
This has been released as part of version 3.1.0 of Amazon.ElasticBeanstalk.Tools. Thanks for the PR and patience to get this out.

PR #70, opened 2019-03-15
Add .NET Core 2.0 deprecation message.

vellozzi:
Show a message about the upcoming .NET Core 2.0 deprecation in Lambda if the function being packaged uses .NET Core 2.0.


src/Amazon.Lambda.Tools/LambdaPackager.cs
```diff
@@ -661,5 +663,36 @@ private static void BundleWithZipCLI(string zipCLI, string zipArchivePath, strin
                 }
             }
         }
+
+        internal static void LogDotNetCore20DeprecationMessageIfNecessary(IToolLogger logger, string targetFramework)
```
normj: Rename the method to LogDeprecationMessageIfNecessary as this method will probably be reused when .NET Core 1.0 is deprecated.

vellozzi: changed (made plural in case they overlap)

PR #77, opened 2019-04-23
Fix for filename or extension is too long issue

boblodgett:
*Issue #, if available:*
73
*Description of changes:*
When a large number of files are to be published to Lambda the build-lambda-zip.exe cannot be started because the list of files exceeds the maximum size of a command. Instead of passing the filenames as command arguments, these changes write the filenames to a temporary file that is then read by build-lambda-zip.


src/Amazon.Lambda.Tools/LambdaPackager.cs
```diff
@@ -553,6 +558,15 @@ private static void BundleWithBuildLambdaZip(string zipArchivePath, string rootD
                     logger?.WriteLine(string.Format("Created publish archive ({0}).", zipArchivePath));
                 }
             }
+            
+            try
```
vellozzi: I'd put the rest of the method in a try-catch with this try-catch in the finally block.
```
try
{
    //lines 515-560
}
finally
{
    try
    {
        File.Delete(inputFilename);
    }
    catch (Exception e)
    {
        logger?.WriteLine($"Warning: Unable to delete temporary input file, {inputFilename}, after zipping files: {e.Message}");
    }  
}
```
That way it will do a best-effort to delete the file no matter what might go wrong during the zip generation.

src/Amazon.Lambda.Tools/BuildLambdaZip/build-lambda-zip.go
```diff
@@ -27,20 +28,37 @@ func main() {
...
+			Usage: "input file path for the list of files to zip.",
+		},
 	}
 
 	app.Action = func(c *cli.Context) error {
-		if !c.Args().Present() {
-			return errors.New("No input provided")
-		}
-
 		firstArg := c.Args().First()
```
jasdel: I don't think the `firstArg` value is valid anymore here since the list of files could instead be within the `input` arg's file.

To maintain the same functionality (line 42) the `outputZip` setting should probably be after line 59 and the first entry in the `files` slice.

src/Amazon.Lambda.Tools/LambdaPackager.cs
boblodgett: Made this change

src/Amazon.Lambda.Tools/BuildLambdaZip/build-lambda-zip.go
boblodgett: Made this change

PR #80, opened 2019-05-05
Add support for request an MFA token code when required.

normj:
Check to see if the AWSCredententials is an assume role credentials and if so provide a callback to request the MFA code from the user.

### DIFF
--- a/src/Amazon.Common.DotNetCli.Tools/AssumeRoleMfaTokenCodeCallback.cs
+++ b/src/Amazon.Common.DotNetCli.Tools/AssumeRoleMfaTokenCodeCallback.cs
@@ -1,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+using Amazon.Runtime;
+
+namespace Amazon.Common.DotNetCli.Tools
+{
+    /// <summary>
+    /// Handle the callback from AWSCredentials when an MFA token code is required.
+    /// </summary>
+    internal class AssumeRoleMfaTokenCodeCallback 
+    {
+        AssumeRoleAWSCredentialsOptions Options { get; set; }
+
+        internal AssumeRoleMfaTokenCodeCallback(AssumeRoleAWSCredentialsOptions options)
+        {
+            this.Options = options;
+        }
+
+        internal string Execute()
+        {
+            Console.Write($"Enter MFA code for {this.Options.MfaSerialNumber}: ");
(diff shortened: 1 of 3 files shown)


src/Amazon.Common.DotNetCli.Tools/AssumeRoleMfaTokenCodeCallback.cs
```diff
@@ -0,0 +1,29 @@
...
+using Amazon.Runtime;
+
+namespace Amazon.Common.DotNetCli.Tools
+{
+    /// <summary>
+    /// Handle the callback from AWSCredentials when an MFA token code is required.
+    /// </summary>
+    internal class AssumeRoleMfaTokenCodeCallback 
+    {
+        AssumeRoleAWSCredentialsOptions Options { get; set; }
```
vellozzi: Could be private.  Doesn't need to be/shouldn't be referenced externally.

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -694,5 +694,33 @@ public static ExecuteShellCommandResult ExecuteShellCommand(string workingDirect
                 return new ExecuteShellCommandResult(proc.ExitCode, capturedOutput.ToString());
             }            
         }
+
+        public static string ReadSecretFromConsole()
```
boblodgett: Is the current behavior of accepting ESC and TAB characters valid? Also, is setting the key to empty the desired effect for just pressing enter or should it check for this case?

src/Amazon.Common.DotNetCli.Tools/AssumeRoleMfaTokenCodeCallback.cs
normj: Updated. Keep in mind though the default is private when a visibility modifier is not specified

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
normj: Updated code to skip the initial ASCII control characters like ESC and TAB

src/Amazon.Common.DotNetCli.Tools/AssumeRoleMfaTokenCodeCallback.cs
vellozzi: Are you sure?  I thought it was internal.
I prefer it to be explicit anyway.

PR #112, opened 2020-03-23
Switch --runtime value to linux-x64

normj:
*Description of changes:*

Change what the value set --runtime when calling dotnet CLI commands for runtimes using Amazon Linux 2.

### DIFF
--- a/src/Amazon.Lambda.Tools/LambdaDotNetCLIWrapper.cs
+++ b/src/Amazon.Lambda.Tools/LambdaDotNetCLIWrapper.cs
@@ -5,6 +5,7 @@ using System.Text;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
 using Amazon.Common.DotNetCli.Tools;
+using System.Linq;
 
 namespace Amazon.Lambda.Tools
 {
@@ -81,7 +82,9 @@ namespace Amazon.Lambda.Tools
             }
 
             arguments.Append($" --manifest \"{fullPackageManifest}\"");
-            arguments.Append($" --runtime {LambdaConstants.RUNTIME_HIERARCHY_STARTING_POINT}");
+
+
+            arguments.Append($" --runtime {LambdaUtilities.DetermineRuntimeParameter(targetFramework)}");
 
             if(!enableOptimization)
             {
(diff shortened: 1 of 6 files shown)


src/Amazon.Lambda.Tools/LambdaDotNetCLIWrapper.cs
```diff
@@ -207,23 +210,38 @@ public int Publish(LambdaToolsDefaults defaults, string projectLocation, string
...
+                    }
+                }
+                else
+                {
                     if (msbuildParameters == null ||
-                        msbuildParameters.IndexOf("--self-contained", StringComparison.InvariantCultureIgnoreCase) == -1)
+                        msbuildParameters.IndexOf("--runtime", StringComparison.InvariantCultureIgnoreCase) == -1)
                     {
-                        arguments.Append(" --self-contained false ");
+                        arguments.Append($" -r {LambdaUtilities.DetermineRuntimeParameter(targetFramework)}");
```
boblodgett: I believe --runtime and -r are the same. Should probably pick one. Also this seems like it could end up duplicating the same thing already added always on line 87.

src/Amazon.Lambda.Tools/LambdaDotNetCLIWrapper.cs
```diff
@@ -207,23 +210,38 @@ public int Publish(LambdaToolsDefaults defaults, string projectLocation, string
...
 
-                // If you set the runtime to RUNTIME_HIERARCHY_STARTING_POINT it will trim out the Windows and Mac OS specific dependencies but Razor view precompilation
-                // will not run. So only do this packaging optimization if there are no Razor views.
-                if (Directory.GetFiles(fullProjectLocation, "*.cshtml", SearchOption.AllDirectories).Length == 0)
+                if (new string[] { "netcoreapp2.0", "netcoreapp2.1" }.Contains(targetFramework))
                 {
-                    arguments.Append($" -r {LambdaConstants.RUNTIME_HIERARCHY_STARTING_POINT}");
+                    if(Directory.GetFiles(fullProjectLocation, "*.cshtml", SearchOption.AllDirectories).Length == 0)
+                    {
+                        arguments.Append($" -r {LambdaConstants.LEGACY_RUNTIME_HIERARCHY_STARTING_POINT}");
```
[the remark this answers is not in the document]
normj: Switch to call LambdaUtilities.DetermineRuntimeParameter and put the setting of runtime and self contained in a shared Action.

src/Amazon.Lambda.Tools/LambdaDotNetCLIWrapper.cs
normj: Standardized on --runtime

PR #129, opened 2020-07-10
Work around packaging issue with Microsoft.PowerShell.SDK

normj:
*Description of changes:*
This PR is fixing PowerShell Lambda functions that use PowerShell 7 `-Parallel` feature which are currently failing because the new runspaces created fail to find the default Modules that should be loaded into.

The reason this is happening is the Lambda bundle is created using the `--runtime linux-x64` switch to reduce bundle size to what is just needed for Linux. This flatten the assemblies in the deployment bundle but the Module files under the runtime folder are still in their original location and so are failed to be found. To work around this issue the publishing tool checks to see if the Microsoft.PowerShell.SDK package is being used and if so move the Modules folder up to the root like the assemblies happened when the package was flatten.

The issue for Microsoft.PowerShell.SDK has been reported on the PowerShell repo which this work around was suggested.
https://github.com/PowerShell/PowerShell/issues/13132

Also noticed time out issues uploading large deployment bundles to S3. Increased the timeout out and switched to use TransferUtility to do multipart uploads.

### DIFF
--- a/src/Amazon.Common.DotNetCli.Tools/Utilities.cs
+++ b/src/Amazon.Common.DotNetCli.Tools/Utilities.cs
@@ -723,5 +723,44 @@ namespace Amazon.Common.DotNetCli.Tools
             }
             return code.ToString().Trim();
         }
+        
+        
+        public static void CopyDirectory(string sourceDirectory, string destinationDirectory, bool copySubDirectories)
+        {
+            // Get the subdirectories for the specified directory.
+            DirectoryInfo dir = new DirectoryInfo(sourceDirectory);
+
+            if (!dir.Exists)
+            {
+                throw new DirectoryNotFoundException(
+                    "Source directory does not exist or could not be found: "
+                    + sourceDirectory);
+            }
+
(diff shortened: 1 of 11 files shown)


src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -723,5 +742,44 @@ public static string ReadSecretFromConsole()
...
+            DirectoryInfo dir = new DirectoryInfo(sourceDirectory);
+
+            if (!dir.Exists)
+            {
+                throw new DirectoryNotFoundException(
+                    "Source directory does not exist or could not be found: "
+                    + sourceDirectory);
+            }
+
+            DirectoryInfo[] dirs = dir.GetDirectories();
```
boblodgett, line 759: This should be moved down into the if (copySubDirectories) block as you don't use it otherwise.

normj: Done

boblodgett (approved):

bpayette (approved):
Looks good to me.

Issue #122, opened 2020-05-07
--msbuild-parameters not working for dotnet lambda package-ci


simon10says:
I've posted my issue at https://github.com/aws/aws-lambda-dotnet/issues/663#issue-613388019 which should be more appropriate here.

simon10says:
Hi @normj , I'm wondering if you have a chance to look at this issue yet?

chrisoverzero:
I am also running into this problem. The reason it's not working is from this code in the Template Processor Manager:

```csharp
// If the project is in the same directory as the CloudFormation template then use any parameters
// there were specified on the command to build the project.
if (IsCurrentDirectory(field.GetLocalPath()))
{
    if (!string.IsNullOrEmpty(this.DefaultOptions.TargetFramework))
        command.TargetFramework = this.DefaultOptions.TargetFramework;
    
    command.Configuration = this.DefaultOptions.Configuration;
    command.DisableVersionCheck = this.DefaultOptions.DisableVersionCheck;
    command.MSBuildParameters = this.DefaultOptions.MSBuildParameters;
}
```

I could easily open an MR to move `command.MSBuildParameters = this.DefaultOptions.MSBuildParameters;` outside of the `if` block, but I'm reluctant to do so because I don't know why it's there. (I don't know what problem this check avoids.) I expect it's an oversight, since the MSBuild parameters are specified as allowed parameters to the `package-ci` command and they're threaded through it in a way that looks intentional. It's only here that there's a mismatch.

chrisoverzero:
Now that I've looked into it more, I'm not sure at all of the benefit of this check. `PackageDotnetProjectAsync` is only called from `ProcessUpdatableResourceAsync`, which is only called from `TransformTemplateAsync`. That's called from `DeployServerlessCommand` and `PackageCICommand`, which correspond to the commands `deploy-serverless` and `package-ci`. Both of these commands are or can be called from the root directory (next to the solution file) and walk the template's defined Functions to their corresponding project files. I'm not at all sure that this is the desired behavior.

An MR can't hurt, so I'll make one as a proposal.

simon10says:
Exploring the MR from @chrisoverzero, I've found a rather quick workaround:

1. Add a `aws-lambda-tools-defaults.json` file in the project folder
2. In the json file, add:
```json
{
  "configuration": "Release",
[...]

PR #132, opened 2020-08-20
Enable MSBuild Parameters for `package-ci` and `deploy-serverless`

chrisoverzero:
These commands accept MSBuild parameters, but ignore them later on.

This also enables the configuration and the verison check disabling.
Not for any deep reason, but only because they were in the same `if`
block. I'm happy to put either of them back in the block if I
misunderstand the purpose.

Closes: #122

*Issue #, if available:*

#122 

*Description of changes:*

The parameters for configuration, version check disabling, and MSBuild parameters were not being passed to the package command when those parameters were supplied to the commands `package-ci` or `deploy-serverless` when that command was called from the root of a project. This removes the locality requirement.

### DIFF
--- a/src/Amazon.Lambda.Tools/TemplateProcessor/TemplateProcessorManager.cs
+++ b/src/Amazon.Lambda.Tools/TemplateProcessor/TemplateProcessorManager.cs
@@ -224,18 +224,12 @@ namespace Amazon.Lambda.Tools.TemplateProcessor
                 LambdaUtilities.DetermineTargetFrameworkFromLambdaRuntime(field.Resource.LambdaRuntime, location);
 
             command.LayerVersionArns = field.Resource.LambdaLayers;
+            if (!string.IsNullOrEmpty(this.DefaultOptions.TargetFramework))
+                command.TargetFramework = this.DefaultOptions.TargetFramework;
 
-            // If the project is in the same directory as the CloudFormation template then use any parameters
-            // there were specified on the command to build the project.
-            if (IsCurrentDirectory(field.GetLocalPath()))
-            {
-                if (!string.IsNullOrEmpty(this.DefaultOptions.TargetFramework))
-                    command.TargetFramework = this.DefaultOptions.TargetFramework;
-                
-                command.Configuration = this.DefaultOptions.Configuration;
-                command.DisableVersionCheck = this.DefaultOptions.DisableVersionCheck;
-                command.MSBuildParameters = this.DefaultOptions.MSBuildParameters;
-            }
+            command.Configuration = this.DefaultOptions.Configuration;
+            command.DisableVersionCheck = this.DefaultOptions.DisableVersionCheck;
+            command.MSBuildParameters = this.DefaultOptions.MSBuildParameters;
                         
             if(!await command.ExecuteAsync())
             {


normj:
Just talking this change through to see if I comfortable with it. 

These commands support deploying multiple projects at the same time through CloudFormation. This is a feature added to these commands which used to only support deploying Functions from a single project. In the old case when there was only one project the msbuild parameters, target framework and configuration made sense for that project. Once multiple projects were added it's not necessarily true those same settings should be applied to all projects. So the assumption was made that if the template was in the same directory as the project for the Lambda/Serverless CloudFormation resource then continue the old behavior. If they were in separate directory then they are using the new feature and rely on settings put in the `aws-lambda-tools-defaults.json` for that particular project. Which I recognize has problems because the settings then have to be locked in a file and not just toggled on a build script command line. 

Does all of that make sense? By making this change we would be saying any settings made on the command line supersede any configurations for all projects via their `aws-lambda-tools-defaults.json` file.

Kralizek:
Maybe a flag to force CLI settings over all projects file could enable the wanted behavior without the risk of overriding project-specific settings.

chrisoverzero:
>Just talking this change through to see if I comfortable with it.

👍 Thank you for the consideration.

> […] rely on settings put in the aws-lambda-tools-defaults.json for that particular project.
> […] have to be locked in a file and not just toggled on a build script command line.

Yup, this is exactly the case which motivated the MR. Now that .NET Core 3.1 has ReadyToRun, `dotnet lambda package-ci` will fail on non-Linux boxes with it activated. We are unable to activate it for "real" build–deployments and disable it for build–deployments which go into our developer sandbox account.

>By making this change we would be saying any settings made on the command line supersede any configurations for all projects via their aws-lambda-tools-defaults.json file.

That is complicating, hrm.

arturomonge:
@normj Thanks for looking into this. I think the problem here is that if there is no `aws-lambda-tools-defaults.json` file in the project's folder, the command line parameters passed to the `package-ci` or `deploy-serverless` commands are still ignored. So in this case:

> any settings made on the command line supersede any configurations for all projects via their aws-lambda-tools-defaults.json file

is not really what's going on because we don't have an `aws-lambda-tools-defaults.json` file to supersede. So maybe the behavior should be: 

_If there **is** a `aws-lambda-tools-defaults.json` file in the project's folder, then ignore the parameters. If there **is not** an `aws-lambda-tools-defaults.json` file in the project's folder, then use the parameters passed to the `package-ci` or `deploy-serverless` commands._

This way, we are not superseding per-project settings but can still use command line parameters that we want to have applied to all projects without a specific configuration. At least, that's what I was expecting: to have those parameters applied to all projects in the template since they did not have a per-project `aws-lambda-tools-defaults.json` file. I did not expect to have these parameters simply ignored. 

Plus, I didn't even know you could have per-project settings, which I only learned when a workaround was proposed [here](https://github.com/aws/aws-extensions-for-dotnet-cli/issues/122#issuecomment-687584379). And even then, it was suggested as a workaround, not as designed behavior. So I think support for multiple projects in terms of the behavior of these parameters should be documented in the `README.md` file or in the commands' help. I can open a PR with the proposed changes to the documentation once the final behavior is defined.

normj:
I merged PR [#152](https://github.com/aws/aws-extensions-for-dotnet-cli/pull/152) that did something very similar to this PR. The difference was in that case you had to explicitly set the parameters on the command to be passed to all of the other publish commands. This one would actually take the values from the defaults file in directory of the template and be applied to all other projects. I think that behavior might be too surprising.

I'm going to close this since I think the other PR solves the issue. Let me know if you disagree.

Issue #113, opened 2020-03-24
.NET Core Lambda Layers for .NET Core 3.1 Broken


normj:
There appears to be a bug in the underlying `dotnet store` command preventing Lambda layers to be created for .NET Core 3.1. This issue is for tracking on the Lambda side but the feature will most likely be blocked till the .NET CLI issue is fixed.

The issue is reported to the dotnet SDK team here: https://github.com/dotnet/sdk/issues/10973

rpopovych:
Does this prevent us from using 3.1 in our lambdas if we're using layers? Can I use a 2.1 layer in 3.1 lambda?

clearwaterstream:
Any update re approach? Wait for MS to fix it or a work-around could be developed in the meantime? @normj  you mentioned

> I have the option to make a modifications to a copy of the inputted manifest and use that when calling dotnet store. I need to come up with a way given a manifest be able to compute what packages to prune or at a minimum detect that package will bring in framework packages and they should upgrade their package versions.

clearwaterstream:
Far from perfect, but the current workaround I use is as follows. Although `dotnet lambda publish-layer` is currently disabled, the AWS Console still allows layer creation :).

What I found is that certain packages will pull a plethora of other packages -- those need to be avoided until a fix is in place. A couple of examples: `Amazon.Lambda.AspNetCoreServer` and `Microsoft.AspNetCore.Authentication.JwtBearer`. Pruning via `PrepforRestoreForComposeStore` works with some but not other packages, I ended up giving up on that workaround.

Let's take this manifest.xml:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <ItemGroup>
    <PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="2.0.0" />
    <PackageReference Include="AWS.Logger.Core" Version="1.6.0" />
    <PackageReference Include="AWSSDK.DynamoDBv2" Version="3.3.105.28" />
    <PackageReference Include="AWSSDK.Lambda" Version="3.3.109.8" />
    <PackageReference Include="AWSSDK.S3" Version="3.3.110.52" />
    <PackageReference Include="AWSSDK.SecretsManager" Version="3.3.102.26" />
    <PackageReference Include="AWSSDK.SimpleSystemsManagement" Version="3.3.119.18" />
    <PackageReference Include="Autofac" Version="5.1.2" />
    <PackageReference Include="FluentValidation" Version="8.6.2" />
[...]

PR #144, opened 2020-10-11
Add support for creating layers with .NET Core 3.1

normj:
*Issue #, if available:*
https://github.com/aws/aws-extensions-for-dotnet-cli/issues/113

*Description of changes:*
For .NET Core the 3.1 the `dotnet store` command used for creating lambda layers will include system dependencies if packages being added to the store also include references to system dependencies. Newtonsoft.Json version 9.0.1 is an example of this. This makes the layer excessively large and breaks prejitting the assemblies in the store.

To fix this the manifest used to create the store used for the layer will now include a list of the system dependencies to prune from the store. In order for this to work it relies on a fix made to the .NET Code SDK version 3.1.400. That thread can be found here. https://github.com/dotnet/sdk/issues/10973


src/Amazon.Lambda.Tools/Commands/PublishLayerCommand.cs
```diff
@@ -156,10 +156,15 @@ private async Task<CreateLayerZipFileResult> CreateRuntimePackageStoreLayerZipFi
...
             {
-                var message = "Publishing runtime package store layers targeting .NET Core 3.1 is currently disabled due to an issue with the dotnet cli's 'store' command " +
-                              "used to create the runtime package store. For further information see the following GitHub issue.\n" +
-                              "https://github.com/dotnet/sdk/issues/10973";
-                throw new LambdaToolsException(message, LambdaToolsException.LambdaErrorCode.DisabledSupportForNET31Layers);
+                var version = DotNetCLIWrapper.GetSdkVersion();
+                
+                // .NET SDK 3.1 versions less then 3.1.400 have an issue throwing NullReferenceExceptions when pruning packages out with the manifest.
+                // https://github.com/dotnet/sdk/issues/10973
+                if (version < Version.Parse("3.1.400"))
```
ppittle, line 163: Consider moving `"3.1.400"` to be a constant in 'LambdaConstants`.

ppittle (approved):

Issue #149, opened 2020-12-08
dotnet lambda deploy-function with an updated framework and runtime doesn't propagate the change


kylegalbraith:
## Description
I just went through and upgraded a bunch of my Lambda functions from `dotnetcore2.1` to `dotnetcore3.1`. However, I have a few projects that use the aws-default config file where you specify the runtime and framework. I updated both of those to be on 3.1 but the function did not get updated to the new version.

## Reproduction Steps
1. Have a project that is on `dotnetcore2.1` published to AWS Lambda and setup `aws-lambda-tools-defaults.json` to be like this:
```
{
    "Information" : [
        "This file provides default values for the deployment wizard inside Visual Studio and the AWS Lambda commands added to the .NET Core CLI.",
        "To learn more about the Lambda commands with the .NET Core CLI execute the following command at the command line in the project root directory.",
        "dotnet lambda help",
        "All the command line options for the Lambda command can be specified in this file."
    ],
    "profile"     : "default",
    "region"      : "us-west-2",
    "configuration" : "Release",
    "framework"     : "netcoreapp2.1",
    "function-runtime" : "dotnetcore2.1",
    "function-memory-size" : 256,
    "function-timeout"     : 60,
    "function-handler"     : "UFSX.AddNewRider::UFSX.AddNewRider.Function::FunctionHandler",
    "function-name"        : "UFSX-AddNewRider",
}
```
2. Upgrade the project to `dotnetcore3.1` and update the json file to have `3.1` for `framework` and `function-runtime`.
3. Run `dotnet lambda deploy-function`

Expected: The `deploy-function` should deploy the function code **and** update the function config to be on `3.1`.
Actual: Deployment works but the runtime is not updated.

P.S. I tried the `update-function-config` as well and that didn't change the runtime either.

ashishdhingra:
Hi @kylegalbraith,

Thanks for reporting the issue. Please confirm if you tried using AWS CLI [update-function-configuration](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/lambda/update-function-configuration.html) and if it works.

Thanks,
Ashish

kylegalbraith:
> Please confirm if you tried using AWS CLI [update-function-configuration](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/lambda/update-function-configuration.html) and if it works.

Yes, this works as expected.

ashishdhingra:
Hi @kylegalbraith,

[...]

PR #169, opened 2021-04-29
Lambda Tool: Fix update issue for runtime and handler

AbubakerB:
*Issue #, if available:*
#149 

*Description of changes:*
When `package-type` is missing from the `aws-lambda-tools-defaults.json` file, any change to runtime or handler is not applied. The change here is to check the existing lambdas package type instead of the defaults file when applying an update.

### DIFF
--- a/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
+++ b/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
@@ -401,8 +401,8 @@ namespace Amazon.Lambda.Tools.Commands
                 different = true;
             }
 
-            var packageType = this.GetStringValueOrDefault(this.PackageType, LambdaDefinedCommandOptions.ARGUMENT_PACKAGE_TYPE, false);
-            if (string.Equals(packageType, Lambda.PackageType.Zip.Value, StringComparison.OrdinalIgnoreCase))
+            var packageType = DeterminePackageType();
+            if (packageType == Lambda.PackageType.Zip)
             {
                 var handler = this.GetStringValueOrDefault(this.Handler, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_HANDLER, false);
                 if (!string.IsNullOrEmpty(handler) && !string.Equals(handler, existingConfiguration.Handler, StringComparison.Ordinal))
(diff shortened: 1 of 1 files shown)


src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -28,6 +28,8 @@ public class UpdateFunctionConfigCommand : LambdaBaseCommand
         {
             LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_NAME,
             LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_DESCRIPTION,
+            LambdaDefinedCommandOptions.ARGUMENT_PACKAGE_TYPE,
```
normj, line 31: Since package type is not updateable as part of the update config API I don't want to expose that as a switch that would make it look like it could be.

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -401,8 +405,8 @@ private UpdateFunctionConfigurationRequest CreateConfigurationRequestIfDifferent
                 different = true;
             }
 
-            var packageType = this.GetStringValueOrDefault(this.PackageType, LambdaDefinedCommandOptions.ARGUMENT_PACKAGE_TYPE, false);
-            if (string.Equals(packageType, Lambda.PackageType.Zip.Value, StringComparison.OrdinalIgnoreCase))
+            var packageType = DeterminePackageType();
```
normj, line 408: Looking at the defaults file is a bug. What the code should be doing is looking at the package type via the `existingConfiguration` object.

normj (changes requested):

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
AbubakerB: Makes sense. Reverted

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
AbubakerB: My initial thoughts were looking at the defaults would be required for the edge case of when a function's package type changes, but after your comment I realised that's not possible.

Updated the code to look at `existingConfiguration`

normj (approved):

normj:
Looks great, thanks. Let get another pair of reviewer eyes on it and then we will get this out.

philasmar (approved):

normj:
Thanks for the PR. It has been released as part of version 5.1.2.

PR #177, opened 2021-06-19
Minor fix to description for dockerfile option

ppejovic:
Dockerfile option description wasn't grammatically correct, have fixed.


normj (approved):

normj:
Thanks for the fix up!

PR #193, opened 2021-11-14
chore: Update tools and tests to at least .NET Core 3.1

normj:
*Description of changes:*
This PR removes the usages of unsupported versions of .NET Core from the build and tests.

* All published tools now target current LTS versions `netcoreapp3.1;net6.0`.
* Remove custom `NETCORE` build config and use standard `NETCOREAPP3_1_OR_GREATER` build config
* Update all test applications to at least .NET Core 3.1
* Remove test applications and tests that were testing specific .NET Core 1.0 behavior.
* Update test projects to .NET 6


ganeshnj (approved):

src/Amazon.Common.DotNetCli.Tools/AbstractCLIWrapper.cs
```diff
@@ -4,7 +4,7 @@
 using System.IO;
 using System.Text;
 
-#if NETCORE
+#if NETCOREAPP3_1_OR_GREATER
```
boblodgett, line 7: Do we need this? Why are we using preprocessor directives NETCOREAPP3_1_OR_GREATER at all when the csproj targetframeworks look like we never build anything below 3.1 anyway?

normj: We need them because the source code is also absorbed by the Visual Studio Toolkit which uses .NET Framework.

boblodgett (approved):

PR #206, opened 2022-03-31
feat: Standardize issue templates for discussions

kellertk:


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,73 @@
...
+    attributes:
+      label: Additional Information/Context
+      description: |
+        Anything else that might be relevant for troubleshooting this bug. Providing context helps us come up with a solution that is most useful in the real world.
+    validations:
+      required: false
+  - type: input
+    id: sdk-version
+    attributes:
+      label: Targeted .NET platform
```
ashovlin, line 65: This doesn't have the package version or some of the other .NET-specific helptext we added to the other repos e.g. https://github.com/aws/aws-lambda-dotnet/blob/68e5159716d3f275087aa33e08dc73c8bf56e7f2/.github/ISSUE_TEMPLATE/bug-report.yml

Is that intentional?

ashishdhingra: I think this was somewhat ported from deleted [.github/ISSUE_TEMPLATE/bug-report.md](https://github.com/aws/aws-extensions-for-dotnet-cli/pull/206/files#diff-5870a741e2fa8d0b03a88d7b036d408ee7e6954051fcc5d7978269bf6d1ff242). Please confirm if we would like to add other fields.

ganeshnj: Copy paste the same placeholder, same for environment

ashishdhingra: Done in bug and feature request templates.

ashovlin: I think for bugs it'd be good to collect the version of the CLI extension that a user is using (e.g. via ` dotnet tool list -g`)

Not sure "Targeted .NET platform" is as relevant here though as it is for other repos:
1. Are you looking for `dotnet --version`?
2. Or, are we really after the target platform of whatever we're trying to deploy? If so, is that covered by reproduction steps

ashishdhingra: I think `dotnet tool list -g` makes more sense for bugs. May be using text area would also help. Should we remove this field from feature request all together?

ganeshnj: include both, tool install and runtime that tool is using is important to debug.

ashishdhingra: Done. Please review.

ganeshnj (approved):

ppittle (approved):

PR #210, opened 2022-04-20
Add ability to enable function url and configure EphemeralStorageSize

normj:
*Description of changes:*
This PR adds following new switches for deploying and updating Lambda functions. They are used to configure function url and the size of the ephemeral storage size.

* `--function-url-enable`
* `--function-url-auth`
* `--ephemerals-storage-size`

The `get-function-config` command was updated to show the new fields.
```
> dotnet lambda get-function-config
Amazon Lambda Tools for .NET Core applications (5.4.0)
Project Home: https://github.com/aws/aws-extensions-for-dotnet-cli, https://github.com/aws/aws-lambda-dotnet

Name:                         TestNewLambdaTool
Arn:                          arn:aws:lambda:us-west-2:123412341234:function:TestNewLambdaTool
Package Type:                 Zip
Runtime:                      dotnet6
Function Handler:             TestNewLambdaTool::TestNewLambdaTool.Functions::Get
Last Modified:                2022-04-20T07:16:20.000+0000
Memory Size:                  256
Ephemeral Storage Size:       512
Role:                         arn:aws:iam::123412341234:role/LambdaDemoRole
Timeout:                      30
Version:                      $LATEST
State:                        Active
Last Update Status:           Successful
KMS Key ARN:                  (default) aws/lambda
Function Url Config
   Url:                       https://3ykyyrgkrbu6t3bovu5eqe2txi0kzbra.lambda-url.us-west-2.on.aws/
   Auth:                      NONE
```

To configure a Lambda Function Url you have to make separate calls CRUD API calls to managed the function url config. If the function auth is set to `NONE` which is the default then permissions need to be added to the function to allow public invoke of the function. This PR manages adding the statement to the function's resource policy when a url config is enabled with `NONE` auth type. It will also remove the statement if the auth type changes to `AWS_IAM` or the function url disabled.

## Bug Fix
Also included in this PR is a fix to on collection properties to for the update function config to be able to clear out the values. That is the reason for the `IsXXXSet` properties set to `true` being added to the `CreateConfigurationRequestIfDifferent` method.

### DIFF
--- a/src/Amazon.Lambda.Tools/Commands/DeployFunctionCommand.cs
+++ b/src/Amazon.Lambda.Tools/Commands/DeployFunctionCommand.cs
@@ -51,6 +51,11 @@ namespace Amazon.Lambda.Tools.Commands
             LambdaDefinedCommandOptions.ARGUMENT_IMAGE_WORKING_DIRECTORY,
             CommonDefinedCommandOptions.ARGUMENT_DOCKER_TAG,
 
+            LambdaDefinedCommandOptions.ARGUMENT_EPHEMERAL_STORAGE_SIZE,
+
+            LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_ENABLE,
+            LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_AUTH,
+
             LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_TAGS,
             LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_SUBNETS,
             LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_SECURITY_GROUPS,
@@ -283,7 +288,16 @@ namespace Amazon.Lambda.Tools.Commands
                         }
                     };
 
-                    if(!string.IsNullOrEmpty(architecture))
+                    var ephemeralSize = this.GetIntValueOrDefault(this.EphemeralStorageSize, LambdaDefinedCommandOptions.ARGUMENT_EPHEMERAL_STORAGE_SIZE, false);
+                    if(ephemeralSize.HasValue)
+                    {
+                        createRequest.EphemeralStorage = new EphemeralStorage
+                        {
+                            Size = ephemeralSize.Value
+                        };
+                    }
+
+                    if (!string.IsNullOrEmpty(architecture))
                     {
                         createRequest.Architectures = new List<string> { architecture };
                     }
@@ -373,6 +387,13 @@ namespace Amazon.Lambda.Tools.Commands
                     catch (Exception e)
                     {
                         throw new LambdaToolsException($"Error creating Lambda function: {e.Message}", LambdaToolsException.LambdaErrorCode.LambdaCreateFunction, e);
+                    }
+
+                    if(this.GetBoolValueOrDefault(this.FunctionUrlEnable, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_ENABLE, false).GetValueOrDefault())
+                    {
+                        var authType = this.GetStringValueOrDefault(this.FunctionUrlAuthType, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_AUTH, false);
+                        await base.CreateFunctionUrlConfig(createRequest.FunctionName, authType);
+                        this.Logger.WriteLine($"Function url config created: {this.FunctionUrlLink}");
                     }
                 }
                 else
@@ -517,6 +538,11 @@ namespace Amazon.Lambda.Tools.Commands
             data.SetIfNotNull(LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_RUNTIME.ConfigFileKey, this.GetStringValueOrDefault(this.Runtime, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_RUNTIME, false));
             data.SetIfNotNull(LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_ARCHITECTURE.ConfigFileKey, this.GetStringValueOrDefault(this.Architecture, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_ARCHITECTURE, false));
             data.SetIfNotNull(LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_LAYERS.ConfigFileKey, LambdaToolsDefaults.FormatCommaDelimitedList(this.GetStringValuesOrDefault(this.LayerVersionArns, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_LAYERS, false)));
+
+            data.SetIfNotNull(LambdaDefinedCommandOptions.ARGUMENT_EPHEMERAL_STORAGE_SIZE.ConfigFileKey, this.GetIntValueOrDefault(this.EphemeralStorageSize, LambdaDefinedCommandOptions.ARGUMENT_EPHEMERAL_STORAGE_SIZE, false));
+
+            data.SetIfNotNull(LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_ENABLE.ConfigFileKey, this.GetBoolValueOrDefault(this.FunctionUrlEnable, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_ENABLE, false));
+            data.SetIfNotNull(LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_AUTH.ConfigFileKey, this.GetStringValueOrDefault(this.FunctionUrlAuthType, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_AUTH, false));
 
             data.SetIfNotNull(LambdaDefinedCommandOptions.ARGUMENT_IMAGE_ENTRYPOINT.ConfigFileKey, LambdaToolsDefaults.FormatCommaDelimitedList(this.GetStringValuesOrDefault(this.ImageEntryPoint, LambdaDefinedCommandOptions.ARGUMENT_IMAGE_ENTRYPOINT, false)));
             data.SetIfNotNull(LambdaDefinedCommandOptions.ARGUMENT_IMAGE_COMMAND.ConfigFileKey, LambdaToolsDefaults.FormatCommaDelimitedList(this.GetStringValuesOrDefault(this.ImageCommand, LambdaDefinedCommandOptions.ARGUMENT_IMAGE_COMMAND, false)));
--- a/src/Amazon.Lambda.Tools/Commands/GetFunctionConfigCommand.cs
+++ b/src/Amazon.Lambda.Tools/Commands/GetFunctionConfigCommand.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Net;
 using System.Text;
 using System.Threading.Tasks;
 using Amazon.Common.DotNetCli.Tools;
@@ -53,9 +54,10 @@ namespace Amazon.Lambda.Tools.Commands
         {
             GetFunctionConfigurationResponse response;
 
+            var functionName = this.GetStringValueOrDefault(this.FunctionName, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_NAME, true);
             try
             {
-                response = await this.LambdaClient.GetFunctionConfigurationAsync(this.GetStringValueOrDefault(this.FunctionName, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_NAME, true));
+                response = await this.LambdaClient.GetFunctionConfigurationAsync(functionName);
             }
             catch (Exception e)
             {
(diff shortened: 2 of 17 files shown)


src/Amazon.Lambda.Tools/Commands/DeployFunctionCommand.cs
```diff
@@ -374,6 +388,13 @@ protected override async Task<bool> PerformActionAsync()
                     {
                         throw new LambdaToolsException($"Error creating Lambda function: {e.Message}", LambdaToolsException.LambdaErrorCode.LambdaCreateFunction, e);
                     }
+
+                    if(this.GetBoolValueOrDefault(this.FunctionUrlEnable, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_ENABLE, false).GetValueOrDefault())
+                    {
+                        var authType = this.GetStringValueOrDefault(this.FunctionUrlAuthType, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_AUTH, false);
+                        await base.CreateFunctionUrlConfig(createRequest.FunctionName, authType);
```
awschristou, line 395: 1. Consider adding a log entry prior to performing this action
2. Should CreateFunctionUrlConfig have an exception handler, similar to CreateFunctionAsync on line 384?

src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
```diff
@@ -417,5 +417,29 @@ public static class LambdaDefinedCommandOptions
...
+                ValueType = CommandOption.CommandOptionValueType.BoolValue,
+                Description = "Enable function URL."
+            };
+        public static readonly CommandOption ARGUMENT_FUNCTION_URL_AUTH =
+            new CommandOption
+            {
+                Name = "Function Url Auth Type",
+                Switch = "--function-url-auth",
+                ValueType = CommandOption.CommandOptionValueType.StringValue,
+                Description = "The type of authentication that your function URL uses, default value is NONE. Valid values: NONE or AWS_IAM"
```
awschristou, line 442: Can the valid values be data driven through (for example) an SDK enum? This would keep the help text from getting stale if a new enum was added

test/Amazon.Lambda.Tools.Integ.Tests/DeployProjectTests.cs
```diff
@@ -278,5 +278,184 @@ public async Task PackageFunctionAsLocalImageThenDeployWithDifferentECRTag()
...
+            command.FunctionName = functionName;
+            command.Role = await TestHelper.GetTestRoleArnAsync();
+            command.Runtime = "dotnet6";
+            command.FunctionUrlEnable = true;
+            command.DisableInteractive = true;
+
+
+            var created = await command.ExecuteAsync();
+            try
+            {
```
awschristou, line 358: I get that the integration test does a couple of things and then has a final cleanup. Would it be possible to move each of those "couple of things" to separate, smaller private methods, so that it is easier to inspect what is being exercised and validated with each one?

src/Amazon.Lambda.Tools/Commands/GetFunctionConfigCommand.cs
```diff
@@ -130,9 +138,33 @@ protected override async Task<bool> PerformActionAsync()
...
+        {
+            try
+            {
+                var urlConfig = (await this.LambdaClient.GetFunctionUrlConfigAsync(new GetFunctionUrlConfigRequest { FunctionName = functionName }));
+                return urlConfig;
+            }
+            catch (AmazonLambdaException ex) when (ex.StatusCode == HttpStatusCode.NotFound || ex.StatusCode == HttpStatusCode.Unauthorized)
+            {
+                return null;
+            }
```
ganeshnj, line 162: neat

src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
```diff
@@ -417,5 +417,29 @@ public static class LambdaDefinedCommandOptions
                 ValueType = CommandOption.CommandOptionValueType.StringValue,
                 Description = "Docker image name and tag in the 'name:tag' format."
             };
+        public static readonly CommandOption ARGUMENT_EPHEMERAL_STORAGE_SIZE =
+            new CommandOption
+            {
+                Name = "Ephemerals Storage Size",
+                Switch = "--ephemerals-storage-size",
+                ValueType = CommandOption.CommandOptionValueType.IntValue,
+                Description = "The size of the function's /tmp directory in MB."
```
ganeshnj, line 426:
```suggestion
                Description = "The size of the function’s /tmp directory in MB. The default value is 512, but can be any whole number between 512 and 10240 MB."
```

src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
```diff
@@ -417,5 +417,29 @@ public static class LambdaDefinedCommandOptions
                 ValueType = CommandOption.CommandOptionValueType.StringValue,
                 Description = "Docker image name and tag in the 'name:tag' format."
             };
+        public static readonly CommandOption ARGUMENT_EPHEMERAL_STORAGE_SIZE =
+            new CommandOption
+            {
+                Name = "Ephemerals Storage Size",
+                Switch = "--ephemerals-storage-size",
```
ganeshnj, line 424: Also should this match CLI parameter `--ephemeral-storage` like we do for all parameter ?

https://docs.aws.amazon.com/cli/latest/reference/lambda/create-function.html

src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
```diff
@@ -417,5 +417,29 @@ public static class LambdaDefinedCommandOptions
...
+                ValueType = CommandOption.CommandOptionValueType.IntValue,
+                Description = "The size of the function's /tmp directory in MB."
+            };
+        public static readonly CommandOption ARGUMENT_FUNCTION_URL_ENABLE =
+            new CommandOption
+            {
+                Name = "Function Url Enable",
+                Switch = "--function-url-enable",
+                ValueType = CommandOption.CommandOptionValueType.BoolValue,
+                Description = "Enable function URL."
```
ganeshnj, line 434:
```suggestion
                Description = "Enable function URL. A function URL is a dedicated HTTP(S) endpoint for your Lambda function. When enabled, Lambda automatically generates a unique URL endpoint for you. Function URL endpoints have the following format: https://<url-id>.lambda-url.<region>.on.aws"
```

From https://docs.aws.amazon.com/lambda/latest/dg/lambda-urls.html

src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
```diff
@@ -417,5 +417,29 @@ public static class LambdaDefinedCommandOptions
...
+            };
+        public static readonly CommandOption ARGUMENT_FUNCTION_URL_AUTH =
+            new CommandOption
+            {
+                Name = "Function Url Auth Type",
+                Switch = "--function-url-auth",
```
ganeshnj, line 440: There are other configs as well for Url property such as Cors, Qualifier. Cors can be interesting one to support. Are these left intentionally?

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-url.html

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -547,7 +600,152 @@ private bool AreDifferent(IEnumerable<string> source, IEnumerable<string> target
...
+
+            try
+            {
+                this.Logger.WriteLine("Adding Lambda permission statement to public access for Function Url");
+                await LambdaClient.AddPermissionAsync(request);
+            }
+            catch(Amazon.Lambda.Model.ResourceConflictException)
+            {
+                this.Logger.WriteLine($"Lambda permission with statement id {LambdaConstants.FUNCTION_URL_PUBLIC_PERMISSION_STATEMENT_ID} for public access already exists");
+            }
```
ganeshnj, line 727: I see at many places exceptions are completely hidden from customer, is there way like passing to enable more logging for troubleshooting such as `--diagnostics` flag?

ganeshnj (changes requested):

src/Amazon.Lambda.Tools/Commands/DeployFunctionCommand.cs
normj: `CreateFunctionAsync` is an SDK call which is why it has the exception handler. `CreateFunctionUrlConfig` is a method in this library and inside that method it makes an SDK call and has exception handler around it. The idea is any exception that comes from the SDK is "expected" and return back a context error message not just the SDK error message. If the logic inside `CreateFunctionUrlConfig` triggers an exception besides the SDK call then it is unexpected and will get caught but the catch block at the start of the error message that will print more diagnostic information for the unexpect error message.

normj: Normally I would put the logging message before the call but since I want to print something the new URL I need the logging message afterwards. I could also put a logging message before hand but that seems extra verbose. If there is an error creating the url the user will get an error message from the exception thrown in `CreateFunctionUrlConfig` that says the function url could not be created.

src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
normj: I switched the code to use the constant values from the SDK but we don't have a mechanism to easily iterate through the available values. If new values are added to the list it is likely we would have to make code changes to handle the function resource policy.

src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
normj: Added

src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
normj: EphemeralStorage is an object so for the CLI when you use `--ephemeral-storage` you have to pass in JSON. This till doesn't have the concept of passing in JSON for properties nor do I want to make users make JSON for just setting the size. What we usually do in this tool is flatten the parameters to make it easier to set. If the is another property added to EphemeralStorage we would most likely add another top level switch.

src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
normj: I added the `A function URL is a dedicated HTTP(S) endpoint for your Lambda function.` part. If I make this too long the formatting looks really bad when users do `dotnet lambda deploy-function --help`.

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
normj: The error messages are not hidden from the user. The Logger is the tools way of writing to either the console windows or the output pane in Visual Studio.

test/Amazon.Lambda.Tools.Integ.Tests/DeployProjectTests.cs
normj: Cleanup up the integ test and added some test utility methods.

src/Amazon.Lambda.Tools/Commands/DeployFunctionCommand.cs
awschristou: Ok -- when I read this, I had thought that CreateFunctionUrlConfig was an SDK method, so your explanation makes sense to me

test/Amazon.Lambda.Tools.Integ.Tests/DeployProjectTests.cs
```diff
@@ -278,5 +278,153 @@ public async Task PackageFunctionAsLocalImageThenDeployWithDifferentECRTag()
...
+                else
+                {
+                    try
+                    {
+                        var policy = (await lambdaClient.GetPolicyAsync(new GetPolicyRequest { FunctionName = functionName })).Policy;
+                        Assert.DoesNotContain(LambdaConstants.FUNCTION_URL_PUBLIC_PERMISSION_STATEMENT_ID, policy);
+                    }
+                    catch (ResourceNotFoundException)
+                    {
+                        // If the last statement is deleted from the policy then a ResourceNotFoundException is thrown which is also proof the statement id has been removed.
```
awschristou, line 364: Is your test guaranteed to always throw? You could replace this try catch with something like `Assert.ThrowsAsync<ResourceNotFoundException>( fnCall )`.

test/Amazon.Lambda.Tools.Integ.Tests/DeployProjectTests.cs
awschristou: Looks great!

awschristou (approved):

test/Amazon.Lambda.Tools.Integ.Tests/DeployProjectTests.cs
normj: What I'm asserting is the permission statement has been removed. If being removed meant the last statement was removed then you get the `ResourceNotFoundException` exception. In my test currently that is always the case but I want to leave it flexible in case things change in the future where the public statement we added was removed but others might exist.

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
ganeshnj: I mean exception here, this is totally hidden from the customer rather they get a customized message. This is from my past experience with deploy tool when not getting extra diagnostic info, which is helpful to diagnose the underlying issues.

normj: This is an unfortunate control flow by exception. I can tell if the permission is already added if I get this exception. Otherwise I have to make another service call to get the resource policy and then parse the JSON to find the statement id then make the call. This keeps the operation to one service call.

ganeshnj (approved):

PR #215, opened 2022-05-13
Fixed issue with Lambda Function URL being accidentally being cleared out

normj:
*Description of changes:*
If a Lambda function currently has a function url configured and then calls deploy without setting the `--function-url-enable` switch the tool would remove the url configuration which it should not have done. This PR checks when the `--function-url-enable`  is not set and uses the current state of the function url to determine if url changes should be made.

### DIFF
--- a/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
+++ b/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
@@ -248,7 +248,13 @@ namespace Amazon.Lambda.Tools.Commands
             }
 
             var urlConfig = await this.GetFunctionUrlConfig(existingConfiguration.FunctionName);
-            var enableUrlConfig = this.GetBoolValueOrDefault(this.FunctionUrlEnable, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_ENABLE, false).GetValueOrDefault();
+
+            // To determine what is the state of the function url check to see if the user explicitly set a value. If they did set a value then use that 
+            // to either add or remove the url config. If the user didn't set a value check to see if there is an existing config to make sure we don't remove
+            // the config url if the user didn't set a value.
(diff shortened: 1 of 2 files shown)


src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -248,7 +248,13 @@ protected async Task UpdateConfigAsync(GetFunctionConfigurationResponse existing
             }
 
             var urlConfig = await this.GetFunctionUrlConfig(existingConfiguration.FunctionName);
-            var enableUrlConfig = this.GetBoolValueOrDefault(this.FunctionUrlEnable, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_ENABLE, false).GetValueOrDefault();
+
+            // To determine what is the state of the function url check to see if the user explicitly set a value. If they did set a value then use that 
+            // to either add or remove the url config. If the user didn't set a value check to see if there is an existing config to make sure we don't remove
+            // the config url if the user didn't set a value.
+            bool enableUrlConfig = this.GetBoolValueOrDefault(this.FunctionUrlEnable, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_ENABLE, false).HasValue ? 
+                                        this.GetBoolValueOrDefault(this.FunctionUrlEnable, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_ENABLE, false).Value : urlConfig != null;
```
ganeshnj, line 256:
```suggestion
            var functionUrlEnableArg = this.GetBoolValueOrDefault(this.FunctionUrlEnable, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_URL_ENABLE, false);
            bool enableUrlConfig = functionUrlEnableArg.HasValue ? functionUrlEnableArg.Value : urlConfig != null;
```

ganeshnj (approved):

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
normj: That is how I had it originally but my concern was having 2 variables `functionUrlEnableArg (bool?)` and `enableUrlConfig (bool)` with such similar names and basically the same datatype would get miss used in the future.

Issue #198, opened 2021-12-20
Incorrect options for push-image in 5.2.0


starkcolin:
## Description
The Docker-specific options are missing from the push-image command in 5.2.0

## Reproduction Steps
Output from dotnet lambda push-image --help in 5.2.0:
```
Amazon Lambda Tools for .NET Core applications (5.2.0)
Project Home: https://github.com/aws/aws-extensions-for-dotnet-cli, https://github.com/aws/aws-lambda-dotnet

push-image:
   Build Lambda Docker image and push the image to Amazon ECR.

   dotnet  lambda push-image [options]
   Options:
      --profile-location                      Optional override to the search location for Profiles, points at a shared credentials file.
      --aws-access-key-id                     The AWS access key id. Used when setting credentials explicitly instead of using --profile.
      --aws-secret-key                        The AWS secret key. Used when setting credentials explicitly instead of using --profile.
      --aws-session-token                     The AWS session token. Used when setting credentials explicitly instead of using --profile.
      -pl    | --project-location             The location of the project, if not set the current directory will be assumed.
      -cfg   | --config-file                  Configuration file storing default values for command line arguments.
      -pcfg  | --persist-config-file          If true the arguments used for a successful deployment are persisted to a config file.
      -farch | --function-architecture        The architecture of the Lambda function. Valid values: x86_64 or arm64. Default is x86_64
      --disable-interactive                   When set to true missing required parameters will not be prompted for.
      --region                                The region to connect to AWS services, if not set region will be detected from the environment.
      --profile                               Profile to use to look up AWS credentials, if not set environment credentials will be used.
      --profile-location                      Optional override to the search location for Profiles, points at a shared credentials file.
      --aws-access-key-id                     The AWS access key id. Used when setting credentials explicitly instead of using --profile.
      --aws-secret-key                        The AWS secret key. Used when setting credentials explicitly instead of using --profile.
      --aws-session-token                     The AWS session token. Used when setting credentials explicitly instead of using --profile.
      -pl    | --project-location             The location of the project, if not set the current directory will be assumed.
      -cfg   | --config-file                  Configuration file storing default values for command line arguments.
      -pcfg  | --persist-config-file          If true the arguments used for a successful deployment are persisted to a config file.
```

Output from 5.1.4:
```
Amazon Lambda Tools for .NET Core applications (5.1.4)
[...]

PR #218, opened 2022-05-24
Fixed an issue where lambda push-image command was ignoring Docker options.

ashishdhingra:
*Issue #, if available:* #198

*Description of changes:*
Fixed an issue where lambda push-image command was ignoring Docker options.
___


src/Amazon.Common.DotNetCli.Tools/Commands/BaseCommand.cs
```diff
@@ -113,6 +113,23 @@ protected static IList<CommandOption> BuildLineOptions(params IList<CommandOptio
             return list;
         }
 
+        /// <summary>
+        /// Used to combine the command specific command options into one list.
+        /// </summary>
+        /// <param name="optionCollections"></param>
+        /// <returns></returns>
+        protected static IList<CommandOption> MergeCommandOptions(params IList<CommandOption>[] optionCollections)
```
normj, line 121: I don't want another method that does 9X% the same as `BuildLineOptions`. Lets modify `BuildLineOptions` to only add things to the union list if it isn't already there. Something like
```
protected static IList<CommandOption> BuildLineOptions(params IList<CommandOption>[] optionCollections)
{
    var list = new List<CommandOption>();
    list.AddRange(CommonOptions);
    if(optionCollections != null)
    {
        foreach (var options in optionCollections)
        {
            foreach(var option in options)
            {
                if (!list.Any(x => string.Equals(x.Switch, option.Switch)))
                {
                    list.Add(option);
                }
            }
        }
    }

    return list;
}
```

Then be sure in the `PushDockerImageCommand` we pass up `BasePushDockerImageCommand<LambdaToolsDefaults>.CommandOptions` instead of `CommonOptions`.

ashishdhingra: Done. Please review.

normj (approved):

ashovlin (approved):

PR #221, opened 2022-06-02
bug: Continue gracefully if unable to get Function Url

ppittle:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/1188

*Description of changes:*
Continue gracefully if unable to get Function Url when updating a Lambda.  Getting Function Url can occur in regions that do not yet have this functionality or if the current context does not have the appropriate permissions to query function urls.

### DIFF
--- a/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
+++ b/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
@@ -247,35 +247,40 @@ namespace Amazon.Lambda.Tools.Commands
                 }
             }
 
-            var urlConfig = await this.GetFunctionUrlConfig(existingConfiguration.FunctionName);
-
-            // To determine what is the state of the function url check to see if the user explicitly set a value. If they did set a value then use that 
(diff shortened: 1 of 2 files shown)


src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -247,8 +248,17 @@ protected async Task UpdateConfigAsync(GetFunctionConfigurationResponse existing
...
+            GetFunctionUrlConfigResponse urlConfig = null;
 
+            try
+            {
+                urlConfig = await GetFunctionUrlConfig(existingConfiguration.FunctionName);
+            }
+            catch (Exception e)
+            {
+                Logger?.WriteLine($"Exception in [{nameof(GetFunctionUrlConfig)}]: {e.Message}{Environment.NewLine}{e.StackTrace}");
+            }
```
ganeshnj, line 260: Do you know which type of exception is thrown when FunctionURL is not available in the region?

Hiding all the errors will lead to bad experience and non intuitive debugging.

ganeshnj (changes requested):

normj (approved):

normj (approved):

PR #242, opened 2022-09-27
chore: Remove S3 buckets which are not used in test applications

dscpinheiro:
*Description of changes:* Remove S3 buckets which are missing encryption and versioning settings, but are not used at all in the test applications.

### DIFF
--- a/testapps/ServerlessWithYamlFunction/serverless.yaml
+++ b/testapps/ServerlessWithYamlFunction/serverless.yaml
@@ -29,9 +29,6 @@ Resources:
             Path: "/"
             Method: GET
 
-  UploadsBucket:
-    Type: AWS::S3::Bucket
-
 Outputs:
   ApiURL:
     Description: API endpoint URL for Prod environment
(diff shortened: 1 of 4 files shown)


ppittle:
Just to triple check, will the Test Apps work if they don't have Upload Buckets?

Curious, would it be better to just turn on Encryption and Versioning for these buckets?

Something like
```
UploadsBucket:
    Type: AWS::S3::Bucket
    Properties:
       BucketEncryption:
          ServerSideEncryptionConfiguration:
               "EncryptionRole"
       VersioningConfiguration:
           Status: Enabled
```

dscpinheiro:
> Just to triple check, will the Test Apps work if they don't have Upload Buckets?
> 
> Curious, would it be better to just turn on Encryption and Versioning for these buckets?
> 
> Something like
> 
> ```yaml
> UploadsBucket:
>     Type: AWS::S3::Bucket
>     Properties:
>        BucketEncryption:
>           ServerSideEncryptionConfiguration:
>                "EncryptionRole"
>        VersioningConfiguration:
>            Status: Enabled
> ```

Yes, they should work (the buckets are not referenced in the test apps).

ppittle (approved):

normj (approved):
I'm baffled why these were ever in here.

PR #244, opened 2022-10-26
Add --resolve-s3 to automatically resolve which S3 bucket to upload build output.

normj:
*Description of changes:*
Similar to the SAM CLI this PR adds a new `--resolve-s3` switch to the `deploy-serverless` and -package-ci` commands. This new switch will allow users to who set this new switch to `true` to have the S3 bucket automatically created and used for uploading deployment bundles.

The default S3 bucket name format is `aws-dotnet-lambda-tools-<region>-<account-id>`.

### DIFF
--- a/src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
+++ b/src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
@@ -206,6 +206,15 @@ namespace Amazon.Lambda.Tools
                 ValueType = CommandOption.CommandOptionValueType.StringValue,
                 Description = "KMS Key ARN of a customer key used to encrypt the function's environment variables"
             };
+        public static readonly CommandOption ARGUMENT_RESOLVE_S3 =
+            new CommandOption
+            {
+                Name = "Resolve S3 Bucket",
+                ShortSwitch = "-rs",
+                Switch = "--resolve-s3",
+                ValueType = CommandOption.CommandOptionValueType.BoolValue,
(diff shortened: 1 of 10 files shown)


src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
```diff
@@ -206,6 +206,15 @@ public static class LambdaDefinedCommandOptions
...
                 Description = "KMS Key ARN of a customer key used to encrypt the function's environment variables"
             };
+        public static readonly CommandOption ARGUMENT_RESOLVE_S3 =
+            new CommandOption
+            {
+                Name = "Resolve S3 Bucket",
+                ShortSwitch = "-rs",
+                Switch = "--resolve-s3",
+                ValueType = CommandOption.CommandOptionValueType.BoolValue,
+                Description = $"If set to true a bucket with the name format of \"{LambdaConstants.DEFAULT_BUCKET_NAME_PREFIX}-<region>-<account-id>\" will be configured to store build outputs"
```
boblodgett, line 216: DEFAULT_BUCKET_NAME_PREFIX ends with a `-` so this will end up with `--` before region. This is incorrect where the actual code below is correct.

boblodgett (changes requested):

src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs
normj: Fixed

boblodgett (approved):

96malhar (approved):

Issue #256, opened 2022-11-29
Unable to use dotnet lambda deploy-function in Linux for AOT


jasonterando:
### Describe the bug

When executing `dotnet lambda deploy-function`  in Linux as a non-root user, it executes Docker, which creates directories using the **root** account.  Subsequently the .zip command fails (outside of the Docker container) because the current user does not have write to the bin/Release folder, which was created by the Docker's root user.

As a workaround, I could install dotnet and execute builds as root, but for security considerations, this isn't ideal.

### Expected Behavior

`dotnet lambda deploy-function` should run successfully without error as a non-root user

### Current Behavior

Trail of tears:

```sh
$ dotnet lambda deploy-function
Amazon Lambda Tools for .NET Core applications (5.6.2)
Project Home: https://github.com/aws/aws-extensions-for-dotnet-cli, https://github.com/aws/aws-lambda-dotnet

Found /etc/os-release
Architecture not provided, defaulting to x86_64 for container build image.
Executing publish command
Starting container for native AOT build using build image: public.ecr.aws/sam/build-dotnet7:latest-x86_64.
... invoking 'docker run --rm --volume "/tmp/Temp-e74a018c-de14-4796-ac6e-aa23bbed068e/LambdaNativeAot/src/LambdaNativeAot":/tmp/source/ --name tempLambdaBuildContainer-6c5d3da2-44a5-49e2-81d5-696f07f3a64a -i public.ecr.aws/sam/build-dotnet7:latest-x86_64 dotnet publish "/tmp/source/" --output "/tmp/source/bin/Release/net7.0/publish" --configuration "Release" --framework "net7.0" --self-contained true /p:GenerateRuntimeConfigurationFiles=true --runtime linux-x64 /p:StripSymbols=true' from directory /tmp/Temp-e74a018c-de14-4796-ac6e-aa23bbed068e/LambdaNativeAot/src/LambdaNativeAot
... docker run: MSBuild version 17.4.0+18d5aef85 for .NET
... docker run:   Determining projects to restore...
... docker run:   Restored /tmp/source/LambdaNativeAot.csproj (in 20.21 sec).
... docker run:   LambdaNativeAot -> /tmp/source/bin/Release/net7.0/linux-x64/bootstrap.dll
... docker run:   LambdaNativeAot -> /tmp/source/bin/Release/net7.0/publish/
Missing deps.json file. Skipping flattening runtime folder because  is an unrecognized format
[...]

PR #257, opened 2022-12-16
Update "lambda package" to use non-root user when launching Docker on Linux/MacOS

jasonterando:
This is related to [Issue #256](https://github.com/aws/aws-extensions-for-dotnet-cli/issues/256)

Description of Change:  Update `dotnet lambda package` so that it launches Docker as the logged in user when being run in Linux or a MacOS.  This will allow `dotnet lambda deploy-function` to successfully create the .zip package on Linux.  This change also has the benfit of not leaving directories around that users have to `chown` or login in as root to clean up.  This will allow non-root developers to successfully use `dotnet lambda` to package and deploy.

Recap of Issue:  When `dotnet lambda package` today builds a project for linux-x64, it launches a Docker container.  Within the container, commands are executed which create /obj and  /bin directories using the Docker container's effective user, which is "root".  When deploying the package, `dotnet lambda deploy-function` attempts to build a .zip within these created directories, and fails, unless the active user is logged in as "root", which is not a good thing to do when you are developing.

Implementation Details:

1. Add class **PosixUserHelper** to assist with running `id` to get the user's User ID and Group ID (unit test **PosixUserHelperTest.cs**)
2. Update **DockerCLIWrapper** to call PosixUserHelper to determine if running under Unix/MacOS and, if so, add "-u" parameter to specify user and group IDs.  Also, change the default locations of .NET and .NuGet to /tmp so that non-root users can install
3. Update **DockerCLIWrapper** to build a list of arguments using List<string> instead of individual variables that are concatenated together.  This will make it easier in the future to add new arguments

### DIFF
--- a/src/Amazon.Common.DotNetCli.Tools/DefaultConfigFile.cs
+++ b/src/Amazon.Common.DotNetCli.Tools/DefaultConfigFile.cs
@@ -2,7 +2,6 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using System.Text;
 using ThirdParty.Json.LitJson;
 
@@ -97,7 +96,7 @@ namespace Amazon.Common.DotNetCli.Tools
                     var obj = new Dictionary<string, string>();
                     foreach (var key in this._rootData[fullSwitchName].PropertyNames)
                     {
-                        obj[key] = this._rootData[fullSwitchName][key]?.ToString();
+                        obj[key] = this._rootData[key]?.ToString();
                     }
                     return obj;
                 }
@@ -134,25 +133,6 @@ namespace Amazon.Common.DotNetCli.Tools
             return null;
         }
 
-        /// <summary>
-        /// Gets the default if it exists as a dictionary.
-        /// </summary>
-        /// <param name="option"></param>
-        /// <returns></returns>
-        public IDictionary<string, string> GetValueAsDictionary(CommandOption option)
-        {
-            var key = option.Switch.Substring(2);
-            var data = this._rootData[key];
-            if (data == null)
-                return null;
-
-            if (data.IsObject)
-            {
-                return data.PropertyNames
-                    .ToDictionary(name => name, name => data[name].ToString());
-            }
-            return null;
-        }
 
         public static string FormatCommaDelimitedList(string[] values)
         {
--- a/src/Amazon.Common.DotNetCli.Tools/DefaultConfigFile.cs
+++ b/src/Amazon.Common.DotNetCli.Tools/DefaultConfigFile.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 using ThirdParty.Json.LitJson;
 
@@ -96,7 +97,7 @@ namespace Amazon.Common.DotNetCli.Tools
                     var obj = new Dictionary<string, string>();
                     foreach (var key in this._rootData[fullSwitchName].PropertyNames)
                     {
-                        obj[key] = this._rootData[key]?.ToString();
+                        obj[key] = this._rootData[fullSwitchName][key]?.ToString();
                     }
                     return obj;
                 }
@@ -133,6 +134,25 @@ namespace Amazon.Common.DotNetCli.Tools
             return null;
         }
 
+        /// <summary>
+        /// Gets the default if it exists as a dictionary.
+        /// </summary>
+        /// <param name="option"></param>
+        /// <returns></returns>
+        public IDictionary<string, string> GetValueAsDictionary(CommandOption option)
+        {
+            var key = option.Switch.Substring(2);
+            var data = this._rootData[key];
+            if (data == null)
+                return null;
+
+            if (data.IsObject)
+            {
+                return data.PropertyNames
+                    .ToDictionary(name => name, name => data[name].ToString());
+            }
+            return null;
+        }
 
         public static string FormatCommaDelimitedList(string[] values)
         {
--- a/src/Amazon.Common.DotNetCli.Tools/DockerCLIWrapper.cs
+++ b/src/Amazon.Common.DotNetCli.Tools/DockerCLIWrapper.cs
@@ -189,18 +189,21 @@ namespace Amazon.Common.DotNetCli.Tools
             if (PosixUserHelper.IsRunningInPosix)
             {
                 var posixUser = PosixUserHelper.GetEffectiveUser(_logger);
-                argumentList.AddRange(new []
+                if (posixUser != null && (posixUser?.UserID > 0 || posixUser.Value.GroupID > 0))
                 {
-                    // Set Docker user and group IDs
-                    "-u",
-                    $"{posixUser.UserID}:{posixUser.GroupID}",
-                    // Set .NET CLI home directory
-                    "-e",
-                    "DOTNET_CLI_HOME=/tmp/dotnet",
-                    // Set NuGet data home directory to non-root directory
-                    "-e",
-                    "XDG_DATA_HOME=/tmp/xdg"
-                });
+                    argumentList.AddRange(new []
+                    {
+                        // Set Docker user and group IDs
+                        "-u",
+                        $"{posixUser.Value.UserID}:{posixUser.Value.GroupID}",
+                        // Set .NET CLI home directory
+                        "-e",
+                        "DOTNET_CLI_HOME=/tmp/dotnet",
+                        // Set NuGet data home directory to non-root directory
+                        "-e",
+                        "XDG_DATA_HOME=/tmp/xdg"
+                    });
+                }
             }
 
             argumentList.Add(imageId);
--- a/src/Amazon.Common.DotNetCli.Tools/DockerCLIWrapper.cs
+++ b/src/Amazon.Common.DotNetCli.Tools/DockerCLIWrapper.cs
@@ -174,7 +174,21 @@ namespace Amazon.Common.DotNetCli.Tools
 
             var imageArg = $"-i {imageId}";
 
-            var arguments = $"run {autoDeleteArg} {mountArg} {containerNameArg} {imageArg} {commandToRun}";
+            // If running in MacOS or Linux, set Docker to launch using the running user's UID and GID,
+            // which will make sure that any directories or files created in mounted volumes will be
+            // writable by the running user's account
+            string userArg;
+            if (PosixUserHelper.IsRunningInPosix)
+            {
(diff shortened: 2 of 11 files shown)


src/Amazon.Common.DotNetCli.Tools/PosixUserHelper.cs
```diff
@@ -0,0 +1,92 @@
...
+                RedirectStandardOutput = true,
+                RedirectStandardError = true,
+                UseShellExecute = false,
+                CreateNoWindow = true,
+            };
+            proc.EnableRaisingEvents = true;
+            proc.OutputDataReceived += (_, e) =>
+            {
+                if (string.IsNullOrEmpty(e.Data)) return;
+                var results = Regex.Match(e.Data, @"uid=(\d+).*gid=(\d+)");
```
Beau-Gosse-dev, line 48: Would it be cleaner and easier to maintain to call `id -u` and `id -g` to get the `uid` and `gid` directly without regex?

src/Amazon.Common.DotNetCli.Tools/PosixUserHelper.cs
```diff
@@ -0,0 +1,92 @@
+using System;
+using System.Diagnostics;
+using System.Text.RegularExpressions;
+
+namespace Amazon.Common.DotNetCli.Tools
+{
+    public struct PosixUserInfo
+    {
+        public uint UserID;
+        public uint GroupID;
```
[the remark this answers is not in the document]
Beau-Gosse-dev: You could also use nullable ints internally instead of checking `uint.MaxValue`. Although extremely unlikely, I guess technically a uid might be the max value.

src/Amazon.Common.DotNetCli.Tools/PosixUserHelper.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            if (userID == uint.MaxValue)
+            {
+                throw new Exception("Unable to get effective user from \"id\"");
+            }
+
+            if (groupID == uint.MaxValue)
+            {
+                throw new Exception("Unable to get effective group from \"id\"");
```
Beau-Gosse-dev, line 82: @normj and team may want you to add to the [ToolExceptions ](https://github.com/aws/aws-extensions-for-dotnet-cli/blob/master/src/Amazon.Common.DotNetCli.Tools/ToolsException.cs#L10) and use those, although I'm not sure if we want to add all of these as different ones. Maybe we can log the details instead and have 1 new exception type related to GetEffectiveUser.

src/Amazon.Common.DotNetCli.Tools/DefaultConfigFile.cs
```diff
@@ -133,6 +134,25 @@ public string GetValueAsString(CommandOption option)
             return null;
         }
 
+        /// <summary>
+        /// Gets the default if it exists as a dictionary.
+        /// </summary>
+        /// <param name="option"></param>
+        /// <returns></returns>
+        public IDictionary<string, string> GetValueAsDictionary(CommandOption option)
```
Beau-Gosse-dev, line 142: This could be a separate PR to the PosixUserInfo change, right?

Beau-Gosse-dev (changes requested):
Thanks for the PR! I tested this with `dotnet lambda package` on Windows against an AOT project and it worked fine.

jasonterando:
Hi, thanks for the feedback, and thanks for confirming it's good on Windows.  If I'm reading correctly, seems like I need to address the following:
1. Add comment to assignment of .NET CLI and NuGet home environment variables:  No problem, will do
2. Change uint to int, maybe make nullable:  Happy to set to int, but did not see Nullable set up in Amazon.Lambda.Tools.csproj, so I think I would have to keep track of whether they are set using a boolean.  Come to think of it, this is probably fine, I can set the defaults for uid and gid to zero (root) and display a warning for Linux or MacOS if I can't use `id` to get the info
3. If there's a preference to do two separate calls to get user and group ID versus a single parsed via regex, that's cool.  Whatever gets the PR approved :)
4. GetValueAsDictionary should have been a separate PR, I must have included this by mistake, will address.

Assuming I do #1, #2 and #4 -- and we decide on #3, is there anything else you need from me to proceed?  Thanks!

Beau-Gosse-dev:
Thanks for the quick follow up @jasonterando! The only thing I would add is that there are still some changes for using dictionary in the CLI args. I didn't look at those changes too closely, but seems like all of those could go in another PR. I'll need one of the repo contributors to review this too, since I don't have write access to this repository.

jasonterando:
Okee doke - I think I _finally_ have the dictionary/parameters thing completely cleared out (I learned my lesson, I will remember next time to not do a PR from local master branch, sorry about the hassle).  Let me know if there is anything else before you feed this up the food chain.

Beau-Gosse-dev:
@normj Is there some from this repo who can review this? I also reproduced the problem on Ubuntu and verified that this fixes it.

src/Amazon.Common.DotNetCli.Tools/DockerCLIWrapper.cs
```diff
@@ -163,18 +164,52 @@ public int Push(string targetTagName)
...
 
-            var imageArg = $"-i {imageId}";
+                "-i"
+            };
+            
+            // For Linux or MacOS, running a .NET image as non-root requires some special
+            // handling.  We have to pass the user's UID and GID, but we also have to specify
+            // the location of where .NET and Nuget put their files, because by default, they
+            // go to the /.dotnet and /.local, which a non-root user can't write to in a container
+            if (PosixUserHelper.IsRunningInPosix)
```
normj, line 189: If the user id is `0` then I think we should skip this part since we are already running at root.

normj (changes requested):

normj:
> @normj Is there some from this repo who can review this? I also reproduced the problem on Ubuntu and verified that this fixes it.

Back from winter holidays and added comments.

src/Amazon.Common.DotNetCli.Tools/PosixUserHelper.cs
```diff
@@ -0,0 +1,107 @@
...
+        public static readonly bool IsRunningInPosix = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ||
+                                                       RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
+
+        /// <summary>
+        /// Return the effective user's UID and GID under Linux and Mac by calling the "id" command.
+        /// This will fault if running on Windows (by design), check IsRunningInPosix before calling this method.
+        /// </summary>
+        /// <returns>PoxUserInfo struct with UID and GID</returns>
+        /// <exception cref="Exception"></exception>
+        public static PosixUserInfo GetEffectiveUser(IToolLogger logger)
```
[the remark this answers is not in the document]
jasonterando: Ok - changing function to return nullable PosixUserInfo and not throw Exceptions

src/Amazon.Common.DotNetCli.Tools/PosixUserHelper.cs
```diff
@@ -0,0 +1,107 @@
...
+        public int UserID;
+        public int GroupID;
+        public bool UserIDSet;
+        public bool GroupIDSet;
+    }
+
+    /// <summary>
+    /// Helper class to retrieve current Linux/Mac user and group information
+    /// </summary>
+    public static class PosixUserHelper
```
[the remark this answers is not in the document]
jasonterando: Acknowledged

src/Amazon.Common.DotNetCli.Tools/DockerCLIWrapper.cs
jasonterando: Ok, changing this to skip if PosixUserHelper isn't able to get the user or group, or if the user and group are both zero (root):

```
                var posixUser = PosixUserHelper.GetEffectiveUser(_logger);
                if (posixUser?.UserID > 0 posixUser?.GroupID > 0)
                {
```

src/Amazon.Common.DotNetCli.Tools/PosixUserHelper.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        /// <summary>
+        /// Return True if running under Unix or Mac
+        /// </summary>
+        public static readonly bool IsRunningInPosix = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ||
+                                                       RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
```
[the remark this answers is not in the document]
jasonterando: Hi - Norm had asked me to make the inclusion of using `System.Runtime.InteropServices` contingent upon NETCOREAPP3_1_OR_GREATER being set, so that this code won't bomb under Visual Studio.  This means that I can't add that as a parameter very easily (OSPlatform is not a c# enum but a "string-in-a-struct" kind of enum).

I was also thinking it wouldn't be a bad idea of separating the "Am I *NIX?" from operations like getting the ID, in case we wanted to add more functionality. 

What I am thinking I should do isto set up an injectable class-based implementation of launching a process and getting results, which would make this testable (we could mock `id` failing or otherwise not being on the system at all).  The same Process situation exists all over this code base, and this implementation could help make that code more testable as well.

jasonterando:
Hi, I've posted an update with changes as follows:

1. Replaced single call to "id" w/ regex parsing to separate calls to "id" for user ID and group ID
2. Failure to call "id" will be logged but ignored
3. If the user is root user and group user no changes will be made to the Docker command
4. Launching the "id" process has been re-implemented using a class to facilitate unit testing (may be reused for other launched processes); added unit tests to simulate "id" success and failure regardless of OS running unit tests
5. References to System.Runtime.InteropServices "using" statements and OS checks are wrapped with "#if NETCOREAPP3_1_OR_GREATER"

Unit and integration test all run ok.  I do not have Visual Studio to confirm things are running OK there.

Beau-Gosse-dev:
FWIW, I tested the latest commit [625bd4f](https://github.com/aws/aws-extensions-for-dotnet-cli/pull/257/commits/625bd4f94020d4ea651f8a994c81070ee0947c4a) on Ubuntu as root and non-root and both worked. When running as root, the -u argument isn't passed, as expected.

normj (approved):

normj:
@jasonterando Nothing more you need to do right now. The change looks good to me and I have run the code through its paces. I need to get somebody else on the team to double check the change and assuming nothing else found we can get the change released.

peterrsongg (approved):
I've tested these changes on a Linux and MacOs instance and it ran as non-root user successfully

normj:
This was released this morning in version 5.6.3 of Amazon.Lambda.Tools

PR #265, opened 2023-03-08
fix: upgrading from EOL .NET version fails in deploy-function command

philasmar:
*Issue #, if available:*
DOTNET-6627

*Description of changes:*
* Lambda configuration in now updated before the function code
* A revert mechanism has been added if function code update fails

### DIFF
--- a/src/Amazon.Lambda.Tools/Commands/DeployFunctionCommand.cs
+++ b/src/Amazon.Lambda.Tools/Commands/DeployFunctionCommand.cs
@@ -464,7 +464,20 @@ namespace Amazon.Lambda.Tools.Commands
                     }
                     else if (packageType == Lambda.PackageType.Image)
                     {
-                        updateCodeRequest.ImageUri = ecrImageUri;                       
+                        updateCodeRequest.ImageUri = ecrImageUri;
+                    }
+
+                    var configUpdated = false;
+                    try
+                    {
+                        // Update config should run before updating the function code to avoid a situation such as
+                        // upgrading from an EOL .NET version to a supported version where the update would fail 
+                        // since lambda thinks we are updating an EOL version instead of upgrading.
+                        configUpdated = await base.UpdateConfigAsync(currentConfiguration, layerPackageInfo.GenerateDotnetSharedStoreValue());
+                    }
+                    catch (Exception e)
+                    {
+                        throw new LambdaToolsException($"Error updating configuration for Lambda function: {e.Message}", LambdaToolsException.LambdaErrorCode.LambdaUpdateFunctionConfiguration, e);
                     }
 
                     try
@@ -473,10 +486,12 @@ namespace Amazon.Lambda.Tools.Commands
                     }
                     catch (Exception e)
                     {
+                        if (configUpdated)
+                        {
+                            await base.AttemptRevertConfigAsync(currentConfiguration);
+                        }
                         throw new LambdaToolsException($"Error updating code for Lambda function: {e.Message}", LambdaToolsException.LambdaErrorCode.LambdaUpdateFunctionCode, e);
                     }
-
-                    await base.UpdateConfigAsync(currentConfiguration, layerPackageInfo.GenerateDotnetSharedStoreValue());
 
                     await base.ApplyTags(currentConfiguration.FunctionArn);
 
--- a/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
+++ b/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
@@ -230,8 +230,28 @@ namespace Amazon.Lambda.Tools.Commands
             }
         }
 
-        protected async Task UpdateConfigAsync(GetFunctionConfigurationResponse existingConfiguration, string dotnetSharedStoreValue)
+        /// <summary>
+        /// Reverts the lambda configuration to an earlier point, which is needed if updating the lambda function code failed.
+        /// </summary>
+        /// <param name="existingConfiguration"></param>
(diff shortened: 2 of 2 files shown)


normj (changes requested):

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -305,6 +325,71 @@ public async Task<GetFunctionConfigurationResponse> GetFunctionConfigurationAsyn
...
+        /// <summary>
+        /// Create an UpdateFunctionConfigurationRequest for the current configuration to be used to revert a failed configuration update.
+        /// </summary>
+        /// <param name="existingConfiguration"></param>
+        /// <returns><see cref="UpdateFunctionConfigurationRequest"/></returns>
+        private UpdateFunctionConfigurationRequest CreateRevertConfigurationRequest(GetFunctionConfigurationResponse existingConfiguration)
+        {
+            var request = new UpdateFunctionConfigurationRequest
+            {
+                FunctionName = existingConfiguration.FunctionName,
```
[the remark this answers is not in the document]
philasmar: It wasn't in the code that we were using to update the config. I didn't seem like a property that can be updated so I left it out to stay consistent.

src/Amazon.Lambda.Tools/Commands/DeployFunctionCommand.cs
```diff
@@ -473,11 +485,10 @@ protected override async Task<bool> PerformActionAsync()
                     }
                     catch (Exception e)
                     {
+                        await base.RevertConfigAsync(currentConfiguration);
```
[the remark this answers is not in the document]
philasmar: Done

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -230,8 +230,29 @@ protected async Task ApplyTags(string functionArn)
             }
         }
 
-        protected async Task UpdateConfigAsync(GetFunctionConfigurationResponse existingConfiguration, string dotnetSharedStoreValue)
+        /// <summary>
+        /// Reverts the lambda configuration to an earlier point, which is needed if updating the lambda function code failed.
+        /// </summary>
+        /// <param name="existingConfiguration"></param>
+        /// <exception cref="LambdaToolsException"></exception>
+        protected async Task RevertConfigAsync(GetFunctionConfigurationResponse existingConfiguration)
```
normj, line 238: Thinking this is only called if there was some error. If this function has an error we probably don't want this method's error being the one that the user sees instead of the original error.

What you do you think about renaming this `AttemptRevertConfigAsync` and any errors are just logged and we catch all exceptions.

normj (changes requested):

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -230,8 +230,29 @@ protected async Task ApplyTags(string functionArn)
...
+            var request = CreateRevertConfigurationRequest(existingConfiguration);
+            await LambdaUtilities.WaitTillFunctionAvailableAsync(Logger, this.LambdaClient, request.FunctionName);
+            this.Logger.WriteLine($"Reverting runtime configuration for function {this.GetStringValueOrDefault(this.FunctionName, LambdaDefinedCommandOptions.ARGUMENT_FUNCTION_NAME, true)}");
+            try
+            {
+                await this.LambdaClient.UpdateFunctionConfigurationAsync(request);
+            }
+            catch (Exception e)
+            {
+                throw new LambdaToolsException($"Error updating configuration for Lambda function: {e.Message}", LambdaToolsException.LambdaErrorCode.LambdaUpdateFunctionConfiguration, e);
```
normj, line 249: I know this will likely change due to my other comment but this error message should say error reverting configuration not updating configuration.

philasmar: Done

normj (approved):

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -305,6 +328,77 @@ public async Task<GetFunctionConfigurationResponse> GetFunctionConfigurationAsyn
             }
         }
 
+        /// <summary>
+        /// Create an UpdateFunctionConfigurationRequest for the current configuration to be used to revert a failed configuration update.
+        /// </summary>
+        /// <param name="existingConfiguration"></param>
+        /// <returns><see cref="UpdateFunctionConfigurationRequest"/></returns>
+        private UpdateFunctionConfigurationRequest CreateRevertConfigurationRequest(GetFunctionConfigurationResponse existingConfiguration)
```
ashovlin, line 336: Comparing this to the API docs, the only gap I see is [FileSystemConfigs](https://docs.aws.amazon.com/lambda/latest/dg/API_UpdateFunctionConfiguration.html#SSS-UpdateFunctionConfiguration-request-FileSystemConfigs), is that okay to omit?

philasmar: This property is not updated by our tooling, so there's no reason to include it in the revert method.

ashovlin (approved):

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
ashovlin: Ah, good point, thanks.

PR #269, opened 2023-04-03
update hosted files endpoint and remove S3 fallback

ghost:
*Description of changes:*

This change updates the file fetcher to pull hosted files from the new, IDE-owned endpoint.
This also removes the S3 Fallback method used for the SDK-owned destinations, because the referenced endpoint is portable, and can be relocated. The previous URL fell back to an S3 bucket in case the CloudFront distribution was ever accidentally deleted - we wouldn't be able to re-create the same URL.


normj (approved):
Change looks good. We also might think about remove the fetcher. It was only used to download some manifest files for .NET Core 2.0 and .NET Core 2.1 which are well out of support.

Issue #264, opened 2023-03-02
"lambda deploy-serverless" command tries to build image when ImageUri is set without any Metadata


roncrush:
### Describe the bug

According to the docs for "AWS::Serverless::Function" the "ImageUri" property can be set if the image has already been pushed to somewhere like ECR (https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-imageuri).

The behavior I'm seeing with the command "lambda deploy-serverless" is that it treats the "ImageUri" like a local path as if it's trying to build the image and push it.

### Expected Behavior

The command should result in the function being successfully deployed by the CloudFormation stack.

### Current Behavior

The following error appears:

> Processing CloudFormation resource \<Function Name\>
> Initiate packaging of \<ImageUri\> for resource <Function Name>
> Directory that the field \<Function Name\>/CodeUri-Or-ImageUri is pointing doesn't exist

### Reproduction Steps

1. Initialize the hello world sample:
```
sam init --package-type Image --base-image amazon/dotnet6-base
```
2. Manually build and deploy the Dockerfile to somewhere like ECR.
3. Update the template.yaml file to look like this (replace the \<ImageUri\> with the deployed image URI):
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: Template to test bug with ImageUri

Globals:
  Function:
    Timeout: 10
    MemorySize: 128

Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      PackageType: Image
      ImageUri: <ImageUri>
```
4, Run the "dotnet lambda deploy-serverless" command (will need a S3 bucket):
```
dotnet lambda deploy-serverless --s3-bucket "some-bucket-for-testing-bug" --s3-prefix "bug/" --stack-name "test-bug" --template "template.yaml"
```

### Possible Solution

It seems like the packaging step should be skipped for Lambda Functions that have an ImageUri set without any metadata.

### Targeted .NET platform

6.0.302

### CLI extension version

| Package Id           |                 Version    |  Commands |
|-------------------|---------------------|----------------------|
| amazon.lambda.tools     |              5.6.3  |      dotnet-lambda |
| dotnet-ef                         |             6.0.8   |     dotnet-ef |

### Environment details (OS name and version, etc.)

macOS Monterey Version 12.6

ashishdhingra:
[...]

PR #272, opened 2023-04-05
Fixed an issue lambda deploy-serverless command tries to build image when ImageUri is set in template without any Metadata.

ashishdhingra:
*Issue #, if available:* #264

*Description of changes:*
Fixed an issue lambda deploy-serverless command tries to build image when ImageUri is set in template without any Metadata.

Per https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-imageuri, `Building your application with necessary Metadata entries takes precedence over ImageUri, so if you specify both then ImageUri is ignored.`.

In customer reported scenario, metadata is not present in YAML (orJSON) template. In such case, it should not try to build the Docker image locally, instead should use the `ImageUri` set in the CloudFormation template. Currently, before this fix, it tries to validate the `ImageUri` as local path and fails the validation check.

**NOTE:** We could have added a validation check to verify if image pointed out by the `ImageUri` exists in ECR, however, there is no API to get image by `ImageUri`.

___

### DIFF
--- a/src/Amazon.Lambda.Tools/TemplateProcessor/ITemplateParser.cs
+++ b/src/Amazon.Lambda.Tools/TemplateProcessor/ITemplateParser.cs
@@ -93,9 +93,9 @@ namespace Amazon.Lambda.Tools.TemplateProcessor
         bool IsCode { get; }
 
         /// <summary>
-        /// True if the field points to an ImageUri.
+        /// True if the field points to an Image already pushed in ECR.
         /// </summary>
-        bool IsImageUri { get; }
+        bool IsImagePushed { get; }
 
 
         /// <summary>
--- a/src/Amazon.Lambda.Tools/TemplateProcessor/ITemplateParser.cs
+++ b/src/Amazon.Lambda.Tools/TemplateProcessor/ITemplateParser.cs
@@ -91,8 +91,13 @@ namespace Amazon.Lambda.Tools.TemplateProcessor
         /// True if the field should contain code like a Lambda package bundle.
         /// </summary>
         bool IsCode { get; }
-        
-        
+
+        /// <summary>
+        /// True if the field points to an ImageUri.
+        /// </summary>
+        bool IsImageUri { get; }
+
+
         /// <summary>
         /// Reference back to the containing updatable resource.
         /// </summary>
--- a/src/Amazon.Lambda.Tools/TemplateProcessor/TemplateProcessorManager.cs
+++ b/src/Amazon.Lambda.Tools/TemplateProcessor/TemplateProcessorManager.cs
@@ -142,12 +142,18 @@ namespace Amazon.Lambda.Tools.TemplateProcessor
             UpdateResourceResults results;
             var localPath = field.GetLocalPath();
 
-            if (!Path.IsPathRooted(localPath))
+            if (!field.IsImageUri && !Path.IsPathRooted(localPath))
                 localPath = Path.Combine(templateDirectory, localPath);
 
             bool deleteArchiveAfterUploaded = false;
+
+            // If ImageUri needs to be processed.
+            if (field.IsImageUri)
+            {
+                results = new UpdateResourceResults { ImageUri = localPath };
+            }
             // Uploading a single file as the code for the resource. If the single file is not a zip file then zip the file first.
-            if(File.Exists(localPath))
+            else if (File.Exists(localPath))
             {
                 if(field.IsCode && !string.Equals(Path.GetExtension(localPath), ".zip", StringComparison.OrdinalIgnoreCase))
                 {
--- a/src/Amazon.Lambda.Tools/TemplateProcessor/TemplateProcessorManager.cs
+++ b/src/Amazon.Lambda.Tools/TemplateProcessor/TemplateProcessorManager.cs
@@ -142,13 +142,13 @@ namespace Amazon.Lambda.Tools.TemplateProcessor
             UpdateResourceResults results;
             var localPath = field.GetLocalPath();
 
-            if (!field.IsImageUri && !Path.IsPathRooted(localPath))
+            if (!field.IsImagePushed && !Path.IsPathRooted(localPath))
                 localPath = Path.Combine(templateDirectory, localPath);
 
             bool deleteArchiveAfterUploaded = false;
 
             // If ImageUri needs to be processed.
-            if (field.IsImageUri)
+            if (field.IsImagePushed)
             {
                 results = new UpdateResourceResults { ImageUri = localPath };
             }
--- a/src/Amazon.Lambda.Tools/TemplateProcessor/UpdatableResource.cs
+++ b/src/Amazon.Lambda.Tools/TemplateProcessor/UpdatableResource.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using System.Text;
+using System.Text.RegularExpressions;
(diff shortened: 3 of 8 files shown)


src/Amazon.Lambda.Tools/TemplateProcessor/TemplateProcessorManager.cs
```diff
@@ -142,12 +142,18 @@ private async Task<UpdateResourceResults> ProcessUpdatableResourceAsync(string t
             UpdateResourceResults results;
             var localPath = field.GetLocalPath();
 
-            if (!Path.IsPathRooted(localPath))
+            if (!field.IsImageUri && !Path.IsPathRooted(localPath))
```
normj, line 145: We don't know if there were users that were relying on the .NET CLI building images with just the `ImageUri` property set. 

We can look to see if the value for ImageUri contains `.dkr.ecr.` and the first token is a numeric. If so then we can assume it is an ECR image and use that otherwise run the local build.

normj (changes requested):
We are going to need a test that we can use a template that has an image uri set to an already pushed image.

src/Amazon.Lambda.Tools/TemplateProcessor/TemplateProcessorManager.cs
ashishdhingra: Added extra check in `IsCode` and `IsImageUri` properties in `UpdatableResource.cs`. Please review.

ashishdhingra:
> We are going to need a test that we can use a template that has an image uri set to an already pushed image.

@normj Added tests for YAML and JSON template.

Could you please advise on the following:
- Possible 2 ways of specifying `ImageUri`:
  - The link https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-imageuri mentions `ImageUri` (and `CodeUri`) as top level template properties.
  - Whereas, https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html specifies `ImageUri` nested inside top level code `Code` property. There is no `CodeUri` property.

  Which one is correct? Or both are supported? If yes, why 2 different ways of specifying `ImageUri`. Should we also test with nested `ImageUri` property template as well?

- During testing, I noticed that the cleanup `DeleteServerlessCommand` when executed without specifying the `Region` (in this case it would use the default region from the profile), even though was successful, but it was not deleting CloudFormation stack. The reason is that we use `TEST_REGION` (`sa-east-1`) while creation of stack and other resources, whereas the `DeleteServerlessCommand` was picking up some other region from my profile. Changing cleanup code to explicitly set `Region` as `TEST_REGION` actually initiated the stack deletion. This cleanup code was copied from existing test. It's highly likely that such existing test(s) might not be actually cleaning (deleting) up the stack. Should we review such cleanup code and correct it as part of this PR?

Thanks,
Ashish

src/Amazon.Lambda.Tools/TemplateProcessor/ITemplateParser.cs
```diff
@@ -91,8 +91,13 @@ public interface IUpdateResourceField
         /// True if the field should contain code like a Lambda package bundle.
         /// </summary>
         bool IsCode { get; }
-        
-        
+
+        /// <summary>
+        /// True if the field points to an ImageUri.
+        /// </summary>
+        bool IsImageUri { get; }
```
normj, line 98: This name keeps confusing me since it is also matching the name of the property. My instinct is to think the property means the `ImageUri` property is set. What it really means is the `ImageUri` is pointing to already pushed image. Can we rename it to something like `IsImagePushed`?

src/Amazon.Lambda.Tools/TemplateProcessor/UpdatableResource.cs
```diff
@@ -140,13 +141,44 @@ public bool IsCode
...
-                    if(!string.IsNullOrEmpty(this._resource.DataSource.GetValue("Code", "ZipFile")))
+                    if (!string.IsNullOrEmpty(this._resource.DataSource.GetValue("Code", "ZipFile")))
                     {
                         // The template contains embedded code.
                         return false;
-                    }                     
+                    }
+                    else
+                    {
+                        string localPath = GetLocalPath();
```
normj, line 151: I find this logic hard to follow. In particular comparing `localPath` with `tempLocalPath` and what is a `tempLocalPath`. What about something like this.

```
string localPath = this._resource.DataSource.GetValueFromResource(LambdaConstants.CF_SERVERLESS_METADATA, LambdaConstants.CF_SERVERLESS_DOCKERCONTEXT);
// Pointing to already pushed image, no package needs to be done.
if (IsECRImage(localPath))
{
    return false;
}
// metadata points to local path that needs to be packaged up return true for code.
else if (!string.IsNullOrEmpty(localPath))
{
    return true;
}

// If no docker metadata was set fallback to looking for a local path in the resource's ImageUri property.
localPath = this._resource.DataSource.GetValue(LambdaConstants.CF_LAMBDA_IMAGEURI);
// Pointing to already pushed image, no package needs to be done.
if (IsECRImage(localPath))
{
    return false;
}

return true;
```

src/Amazon.Lambda.Tools/TemplateProcessor/UpdatableResource.cs
```diff
@@ -140,13 +141,44 @@ public bool IsCode
...
 
-                    return true;
+            public bool IsImageUri
+            {
+                get
+                {
+                    string localPath = GetLocalPath();
+
+                    string tempLocalPath = this._resource.DataSource.GetValueFromResource(LambdaConstants.CF_SERVERLESS_METADATA, LambdaConstants.CF_SERVERLESS_DOCKERCONTEXT);
+                    if (string.IsNullOrEmpty(tempLocalPath))
```
normj, line 175: Same comment as above that the flow and naming is confusing. Maybe you could use the same pattern I suggest above for this section.

src/Amazon.Lambda.Tools/TemplateProcessor/UpdatableResource.cs
```diff
@@ -179,6 +211,11 @@ public string GetMetadataDockerTag()
             {
                 return this.DataSource.GetValueDictionaryFromResource("Metadata", "DockerBuildArgs");
             }
+
+            private bool IsECRImage(string path)
```
normj, line 215: You can make this public and static and then add some unit tests for this method. Since this is a tool and not a library we can be more open to making things public and not worry about changing the API later.

normj (changes requested):

src/Amazon.Lambda.Tools/TemplateProcessor/ITemplateParser.cs
ashishdhingra: Done.

src/Amazon.Lambda.Tools/TemplateProcessor/UpdatableResource.cs
ashishdhingra: Done

src/Amazon.Lambda.Tools/TemplateProcessor/UpdatableResource.cs
ashishdhingra: Done

normj (approved):

96malhar (approved):

Issue #274, opened 2023-04-12
Container builds without self-contained msbuild parameter fail with Unknown error executing command: Value cannot be null. (Parameter 'inputUri')


Beau-Gosse-dev:
### Describe the bug

When running a `dotnet publish` on a container without setting `"msbuild-parameters": "--self-contained true"` in `aws-lambda-tools-defaults.json`, the code tries to look up the project file to look for properties inside it, however at that point, the project file path is for use inside the container at `/tmp/source` so the host process can't find the project file and errors out. 

### Expected Behavior

The code should use the project file path on the host to find the file

### Current Behavior

It uses the container path and errors out since the container isn't started yet.

### Reproduction Steps

try to `dotnet lambda package` a native AOT .NET 7 app without  `"msbuild-parameters": "--self-contained true"` in `aws-lambda-tools-defaults.json`

### Possible Solution

https://github.com/aws/aws-extensions-for-dotnet-cli/pull/275

### Additional Information/Context

Found in this canary which uses SAM CLI: https://github.com/aws-samples/serverless-dotnet-demo/actions/runs/4673160677/jobs/8276085135

### Targeted .NET platform

net7

### CLI extension version

5.6.5

### Environment details (OS name and version, etc.)

windows 10

PR #275, opened 2023-04-12
Don't use container project path for file lookup outside container

Beau-Gosse-dev:
*Issue #, if available:*
https://github.com/aws/aws-extensions-for-dotnet-cli/issues/274

*Description of changes:*
Pass down both the path in the container and on the host so different parts of the code can do what they need to. 

This should also affect this line: https://github.com/Beau-Gosse-dev/aws-extensions-for-dotnet-cli/blob/master/src/Amazon.Lambda.Tools/LambdaDotNetCLIWrapper.cs#L342

This wasn't caught in tests because all tests had msbuild parameters passed, so this part of the code was never hit. This also changes the tests so that it will be hit in 2 cases.

### DIFF
--- a/RELEASE.CHANGELOG.md
+++ b/RELEASE.CHANGELOG.md
@@ -1,3 +1,8 @@
+### Release 2023-04-11
+* **Amazon.Lambda.Tools (5.6.5)**
+  * Allow self contained in csproj.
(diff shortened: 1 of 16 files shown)


Beau-Gosse-dev:
A work around for SAM CLI is here: https://github.com/aws/aws-extensions-for-dotnet-cli/issues/274#issuecomment-1507431721

dscpinheiro (approved):

PR #276, opened 2023-04-19
LambdaUtilities: .NET 7 supported on ARM Amazon Linux 2

dmunch:
The (removed) Exception references the closed issue https://github.com/dotnet/runtime/issues/76195. 

After further investigation, I found that .NET 7 is supported on ARM Amazon Linux 2 since the following PR has been merged and released as .NET 7.0.4 on March 14, 2023: [[release/7.0] Target lower glibc for Linux arm64](https://github.com/dotnet/runtime/pull/80866).

This PR only removes the Exception - however, the referenced Docker image `public.ecr.aws/sam/build-dotnet7:latest-arm64` doesn't exist yet on ECR, which is unfortunately not something I can fix. 

As a workaround, I've build a local docker image on my Mac M1 with the following Dockerfile and the following command, which tags it the way the dotnet lambda tool expects its. 

```sh
docker build . -t "public.ecr.aws/sam/build-dotnet7:latest-arm64"
```

This way, I was able to successfully build a .NET7 AOT Lamba on ARM64 locally, which runs happily on AWS Lambda.

```Dockerfile
FROM public.ecr.aws/amazonlinux/amazonlinux:2 AS base
WORKDIR /source

# Install .NET 7 and other dependencies for compiling natively
RUN yum update -y && yum install -y clang krb5-devel openssl-devel zip libicu-devel wget tar gzip

RUN wget https://dot.net/v1/dotnet-install.sh -O dotnet-install.sh
RUN chmod +x dotnet-install.sh 
RUN ./dotnet-install.sh --channel 7.0 --install-dir /opt/dotnet

ENV DOTNET_ROOT="/opt/dotnet"
ENV PATH="$PATH:/opt/dotnet:/opt/dotnet/tools"
```

### DIFF
--- a/src/Amazon.Lambda.Tools/LambdaUtilities.cs
+++ b/src/Amazon.Lambda.Tools/LambdaUtilities.cs
@@ -118,10 +118,8 @@ namespace Amazon.Lambda.Tools
 
             switch (targetFramework?.ToLower())
             {
-                case TargetFrameworkMonikers.net70:
-                    return architecture == LambdaConstants.ARCHITECTURE_ARM64 ?
-                        throw new LambdaToolsException(".NET 7 is not supported on ARM Amazon Linux 2. See https://github.com/dotnet/runtime/issues/76195", LambdaToolsException.LambdaErrorCode.Net7OnArmNotSupported)
-                        : $"public.ecr.aws/sam/build-dotnet7:latest-{architecture}";
+                case TargetFrameworkMonikers.net70:                    
+                    return $"public.ecr.aws/sam/build-dotnet7:latest-{architecture}";
                 case TargetFrameworkMonikers.net60:
                     return $"public.ecr.aws/sam/build-dotnet6:latest-{architecture}";
                 case TargetFrameworkMonikers.netcoreapp31:


Beau-Gosse-dev:
@mrkdeng

normj:
@dmunch We are waiting on the team that publishes the Lambda build images to publish the .NET 7 ARM image and then we will update the tooling to use it. Hopefully that will be done soon but I know they have a lot of other competing priorities right now.

mrkdeng:
@dmunch Thank you for your contribution! Here is just a reference to the image repo that we've been working on https://github.com/aws/aws-sam-build-images/pull/81. We still don't know when it will be published yet. But let's also double check this change passes all tests. One example is https://github.com/aws/aws-extensions-for-dotnet-cli/blob/560dfb576895feaa8e834fc958d0e0aacb9b917b/test/Amazon.Lambda.Tools.Test/UtilitiesTests.cs#L175

Beau-Gosse-dev:
@normj The .NET 7 ARM SAM build image just went live, so this should be unblocked. 

https://gallery.ecr.aws/sam/build-dotnet7

public.ecr.aws/sam/build-dotnet7:latest-arm64

normj:
Great @Beau-Gosse-dev, I'll do some final verification and then work on getting this released.

normj:
Confirm successful building of an ARM AOT function from M1 Mac

normj (approved):

boblodgett (approved):

normj:
Looks like I'm going to be delayed getting this released because some of our tests are still using .NET Core 3.1 and since .NET Core 3.1 recently hit phase 2 of deprecation, meaning you can't create .NET Core 3.1 functions anymore, the tests are now failing. I'll try and get the test fixed up quickly so we can get this released.

normj:
Support for .NET 7 ARM AOT builds is now enabled with this PR as part of version 5.7.0.

PR #281, opened 2023-06-19
Improve error reporting used by the VS Toolkit

normj:
*Description of changes:*
Rework in the DeployFunctionCommand and DeployServerlessCommand to replace usages of `return false` when there is an error to abort the deployment to throw a `LambdaToolsException` instead. This gives the toolkit better insight what actually went wrong and allows them to display a prompt with the error message.

All exceptions are eventually trapped at the base method called by the CLI or Toolkit so returning exceptions instead of returning false will have no introduce surprise failures.

### DIFF
--- a/src/Amazon.Lambda.Tools/Commands/DeployServerlessCommand.cs
+++ b/src/Amazon.Lambda.Tools/Commands/DeployServerlessCommand.cs
@@ -238,9 +238,7 @@ namespace Amazon.Lambda.Tools.Commands
             // All other states means the Stack is in an inconsistent state.
             else
             {
-                this.Logger.WriteLine($"The stack's current state of {existingStack.StackStatus} is invalid for updating");
-                return false;
-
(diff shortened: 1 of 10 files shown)


src/Amazon.Lambda.Tools/Commands/DeployServerlessCommand.cs
```diff
@@ -337,8 +335,7 @@ protected override async Task<bool> PerformActionAsync()
             }
 
             // The change set can take a few seconds to be reviewed and be ready to be executed.
-            if (!await WaitForChangeSetBeingAvailableAsync(changeSetResponse.Id))
-                return false;
+            await WaitForChangeSetBeingAvailableAsync(changeSetResponse.Id);
```
normj, line 338: The `WaitForChangeSetBeingAvailableAsync` method was refactored to no longer return booleans and instead throw an exception if there was an error.

shruti0085 (approved):
Thank you for the quick turnaround on this!

philasmar (approved):

PR #282, opened 2023-06-26
fix: ensure net6.0 runtimes are copied with the publish artifacts

96malhar:
*Issue #, if available:*
DOTNET-6497
https://github.com/aws/aws-lambda-dotnet/issues/1341

*Description of changes:*
This PR copies the `net6.0` runtime modules in the publish directory while deploying PS Lambda functions based on .NET 6.

The `Microsoft.PowerShell.SDK` version referenced by `TestPowerShellParallelTest.csproj` is also bumped up to `7.2.1` to run the PS Lambda function on .NET 6.

### DIFF
--- a/src/Amazon.Lambda.Tools/LambdaPackager.cs
+++ b/src/Amazon.Lambda.Tools/LambdaPackager.cs
@@ -442,8 +442,7 @@ namespace Amazon.Lambda.Tools
 
         /// <summary>
         /// Work around issues with Microsoft.PowerShell.SDK NuGet package not working correctly when publish with a
-        /// runtime switch. The nested Module folder under runtimes/win/lib/netcoreapp3.1/ needs to be copied
-        /// to the copied to the root of the deployment bundle.
+        /// runtime switch. The nested Module folder under runtimes/unix/lib/{targetFramework}/ needs to be copied to the root of the deployment bundle.
         ///
         /// https://github.com/PowerShell/PowerShell/issues/13132
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/Amazon.Lambda.Tools/LambdaPackager.cs
```diff
@@ -451,7 +451,12 @@ private static bool ValidateDependencies(IToolLogger logger, string targetFramew
         /// <param name="publishLocation"></param>
         private static void FlattenPowerShellRuntimeModules(IToolLogger logger, string publishLocation, string targetFramework)
         {
-            var runtimeModuleDirectory = Path.Combine(publishLocation, "runtimes/win/lib/netcoreapp3.1/Modules");
+            var runtimeModuleDirectory = targetFramework switch
+            {
+                "net6.0" => Path.Combine(publishLocation, "runtimes/unix/lib/net6.0/Modules"),
```
96malhar, line 456: We need to copy the runtimes from the `unix` directory instead of `win` to support deployments from Unix based OSes as well.

src/Amazon.Lambda.Tools/LambdaPackager.cs
```diff
@@ -451,7 +451,12 @@ private static bool ValidateDependencies(IToolLogger logger, string targetFramew
         /// <param name="publishLocation"></param>
         private static void FlattenPowerShellRuntimeModules(IToolLogger logger, string publishLocation, string targetFramework)
         {
-            var runtimeModuleDirectory = Path.Combine(publishLocation, "runtimes/win/lib/netcoreapp3.1/Modules");
+            var runtimeModuleDirectory = targetFramework switch
```
normj, line 454: I don't think we need to have the switch. You can just do `runtimeModuleDirectory = $"runtimes/unix/lib/{targetFramework}/Modules"`. Then we don't have to go back and update this for .NET 8.

normj (changes requested):

src/Amazon.Lambda.Tools/LambdaPackager.cs
96malhar: Fixed

normj (approved):

ashishdhingra (approved):

PR #294, opened 2023-11-12
Enable AOT publishing for AL2023

normj:
*Description of changes:*
The deploy tooling when building for AOT will default to container builds unless we detect we on Amazon Linux. In that case container build defaults to false and we compile on the Amazon Linux instance. Originally the code only checked for AL2 but now that we have AL2023 and .NET 8 managed runtime will be based on this I extended the check for AL 2023.

I also added the .NET 8 target framework to the project file.

I verified the fix by spinning up an EC2 AL2023 instance and deployed an AOT function from there which did not trigger a container build.

### DIFF
--- a/src/Amazon.Lambda.Tools/LambdaPackager.cs
+++ b/src/Amazon.Lambda.Tools/LambdaPackager.cs
@@ -21,6 +21,7 @@ namespace Amazon.Lambda.Tools
         private const string BootstrapFilename = "bootstrap";
         private const string LinuxOSReleaseFile = @"/etc/os-release";
         private const string AmazonLinux2InOSReleaseFile = "PRETTY_NAME=\"Amazon Linux 2\"";
+        private const string AmazonLinux2023InOSReleaseFile = "PRETTY_NAME=\"Amazon Linux 2023\"";
 #if NETCOREAPP3_1_OR_GREATER        
         private static readonly string BuildLambdaZipCliPath = Path.Combine(
             Path.GetDirectoryName(new Uri(Assembly.GetExecutingAssembly().Location).LocalPath),
@@ -36,7 +37,7 @@ namespace Amazon.Lambda.Tools
             { "netcoreapp1.1", Version.Parse("1.6.1") }
         };
 
-        private static bool IsAmazonLinux2(IToolLogger logger)
+        private static bool IsAmazonLinux(IToolLogger logger)
         {
 #if !NETCOREAPP3_1_OR_GREATER
     return false;
@@ -51,6 +52,12 @@ namespace Amazon.Lambda.Tools
                     {
                         logger?.WriteLine(
                             $"Linux distribution is Amazon Linux 2, NativeAOT container build is optional");
+                        return true;
+                    }
+                    if (readText.Contains(AmazonLinux2023InOSReleaseFile))
+                    {
+                        logger?.WriteLine(
+                            $"Linux distribution is Amazon Linux 2023, NativeAOT container build is optional");
                         return true;
                     }
                 }
@@ -92,7 +99,7 @@ namespace Amazon.Lambda.Tools
             else if (string.IsNullOrWhiteSpace(containerImageForBuild))
             {
                 // Use a default container image if Use Container is set to true, or if we need to build NativeAOT on non-AL2
-                if ((useContainerForBuild.HasValue && useContainerForBuild.Value) || (isNativeAot && !IsAmazonLinux2(logger)))
+                if ((useContainerForBuild.HasValue && useContainerForBuild.Value) || (isNativeAot && !IsAmazonLinux(logger)))
                 {
                     containerImageForBuild = LambdaUtilities.GetDefaultBuildImage(targetFramework, architecture, logger);
                 }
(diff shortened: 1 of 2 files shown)


dscpinheiro (approved):

src/Amazon.Lambda.Tools/LambdaPackager.cs
```diff
@@ -36,7 +37,7 @@ public static class LambdaPackager
             { "netcoreapp1.1", Version.Parse("1.6.1") }
         };
 
-        private static bool IsAmazonLinux2(IToolLogger logger)
+        private static bool IsAmazonLinux(IToolLogger logger)
```
Beau-Gosse-dev, line 40: Originally I was thinking this method would need to change into something more complex like `IsCorrectAmazonLinux(IToolLogger logger, string targetFramework)`

Then we can use `targetFramework` to check if `.net6.0` expect `al2` and if `.net8.0` expect `al2023`.

But I guess as long as the customer includes `libicu` in their project file, either targetFramework will work with either Amazon Linux distribution. Is that right?

src/Amazon.Lambda.Tools/LambdaPackager.cs
```diff
@@ -92,7 +99,7 @@ private static bool IsAmazonLinux2(IToolLogger logger)
             else if (string.IsNullOrWhiteSpace(containerImageForBuild))
             {
                 // Use a default container image if Use Container is set to true, or if we need to build NativeAOT on non-AL2
-                if ((useContainerForBuild.HasValue && useContainerForBuild.Value) || (isNativeAot && !IsAmazonLinux2(logger)))
+                if ((useContainerForBuild.HasValue && useContainerForBuild.Value) || (isNativeAot && !IsAmazonLinux(logger)))
```
Beau-Gosse-dev, line 102: I think previously we had the benefit of only having AL2 to deploy to, so we knew what OS we had to build on. But here, what if the customer is using an AL2023 build image (for example, a SAM CLI provided container image that is AL2023), but their Lambda is still configured to deploy to provided.al2? 

Seems like we should either check that the `OS build version = OS runtime version` or we could just say like `.net 6 = al2` and `.net 8 = al2023`. Locking a .NET version to an Amazon Linux version does reduce the testing and support landscape, but limits customer options.

Beau-Gosse-dev: Or maybe AL2 and AL2023 are close enough that they are cross-build compatible?

src/Amazon.Lambda.Tools/LambdaPackager.cs
Beau-Gosse-dev: Talking offline, if you've been running .NET 8 on AL2 without issues, then this isn't important

src/Amazon.Lambda.Tools/LambdaPackager.cs
```diff
@@ -92,7 +99,7 @@ private static bool IsAmazonLinux2(IToolLogger logger)
             else if (string.IsNullOrWhiteSpace(containerImageForBuild))
             {
                 // Use a default container image if Use Container is set to true, or if we need to build NativeAOT on non-AL2
```
Beau-Gosse-dev, line 101: Could update this comment to say non-AmazonLinux

src/Amazon.Lambda.Tools/LambdaPackager.cs
Beau-Gosse-dev: Seems like a rare issue that built on AL2 is deployed to AL2023 or vice versa, and it could be compatible anyway. No need to change it.

src/Amazon.Lambda.Tools/LambdaPackager.cs
Beau-Gosse-dev: I don't have permissions to approve the PR, but this comment is a nit so otherwise, looks good to me.

jamiewinder:
I'm unsure what I'm missing to get this to work. I have ADO pipeline that builds my .NET 6 Lambda and I want to convert this to use .NET 8 with AOT. I've updated the project accordingly and made sure I'm on version 5.9 of these tools.

However, when `dotnet lambda publish` runs, I get the error:

> Found /etc/os-release
> Architecture not provided, defaulting to x86_64 for container build image.
> No container build image available for targetFramework net8.0 and architecture x86_64.

So it appears to be defaulting to a container build which I suppose is what I'm trying to avoid. However the pipeline itself is running on Ubuntu 22. Do I just need to wait for a suitable container build image to be available or is there some configuration I'm missing to enable this now?

normj:
@jamiewinder When building on a non Amazon Linux environment you need to use container build matching Lambda's runtime for AOT. Unfortunately we haven't released a .NET 8 Lambda build image yet. It should be released together with the managed runtime.

Beau-Gosse-dev:
@jamiewinder, in the meantime, you can build your own image to use, or for testing purposes (not production ready) you can use the image that James Eastham built like this: `dotnet lambda deploy-function -cifb plantpowerjames/dotnet-8-lambda-build:al2023-8.0.100`  

(cifb stands for Container Image For Build)

James talks about it more here: https://youtu.be/kyb16r-Oul0?si=Fo9l7sS4xTii0ZEz

PR #307, opened 2024-03-27
fix: use msbuild to retrieve outputtype

philasmar:
*Issue #, if available:*
DOTNET-7400

*Description of changes:*
* Use `dotnet msbuild` to determine the value of the `OutputType` property. This fixes an issue where deploying .NET 8 Native AOT Web Apps fails because of a check that looks for the `OutputType` property explicitly in the .csproj. In .NET8, web apps now inherit the `OutputType` property from the SDK.

### DIFF
--- a/src/Amazon.Common.DotNetCli.Tools/Utilities.cs
+++ b/src/Amazon.Common.DotNetCli.Tools/Utilities.cs
@@ -203,7 +203,7 @@ namespace Amazon.Common.DotNetCli.Tools
         public static string LookupOutputTypeFromProjectFile(string projectLocation)
         {
             var process = new Process();
-            var outputJson = string.Empty;
+            var output = string.Empty;
             var msbuildProcessFailed = false;
             try
             {
@@ -217,7 +217,7 @@ namespace Amazon.Common.DotNetCli.Tools
                 };
 
                 process.Start();
-                outputJson = process.StandardOutput.ReadToEnd();
+                output = process.StandardOutput.ReadToEnd();
                 var hasExited = process.WaitForExit(5000);
 
                 // If it hasn't completed in the specified timeout, stop the process and give up
@@ -244,13 +244,13 @@ namespace Amazon.Common.DotNetCli.Tools
                 var projectFile = FindProjectFileInDirectory(projectLocation);
                 var xdoc = XDocument.Load(projectFile);
                 var element = xdoc.XPathSelectElement("//PropertyGroup/OutputType");
-                outputJson = element?.Value;
+                output = element?.Value;
             }
 
             return 
-                string.IsNullOrEmpty(outputJson) ? 
+                string.IsNullOrEmpty(output) ? 
                     null : 
-                    outputJson.Trim();
+                    output.Trim();
         }
 
         public static bool LookPublishAotFlag(string projectLocation, string msBuildParameters)
--- a/src/Amazon.Common.DotNetCli.Tools/Utilities.cs
+++ b/src/Amazon.Common.DotNetCli.Tools/Utilities.cs
@@ -195,14 +195,67 @@ namespace Amazon.Common.DotNetCli.Tools
             return element?.Value;
         }
 
+        /// <summary>
+        /// Uses `dotnet msbuild` to determine project properties.
+        /// </summary>
+        /// <param name="projectLocation">Path of the project</param>
+        /// <param name="propertyName">The name of the property</param>
+        /// <returns>The value of the project property</returns>
+        private static string LookupMsbuildProperty(string projectLocation, string propertyName)
+        {
+            try
+            {
+                var process = new Process()
+                {
+                    StartInfo = new ProcessStartInfo()
+                    {
+                        FileName = "dotnet",
+                        Arguments = $"msbuild {projectLocation} -getProperty:{propertyName}",
+                        RedirectStandardOutput = true,
+                        UseShellExecute = false,
+                        WindowStyle = ProcessWindowStyle.Hidden
+                    }
+                };
+
+                process.Start();
+                var outputJson = process.StandardOutput.ReadToEnd();
+                var hasExited = process.WaitForExit(5000);
+
+                // If it hasn't completed in the specified timeout, stop the process and give up
+                if (!hasExited) 
+                {
+                    process.Kill();
+                    return null;
+                }
+
+                // If it has completed but unsuccessfully, give up
+                if (process.ExitCode != 0)
+                {
+                    return null;
+                }
+
+                if (string.IsNullOrEmpty(outputJson))
+                {
+                    return null;
+                }
+
+                return outputJson.Trim();
+            }
+            catch (Exception)
+            {
+                // swallow any exceptions related to `dotnet msbuild`
+                return null;
+            }
+        }
+
+        /// <summary>
+        /// Retrieve the `OutputType` property of a given project
+        /// </summary>
+        /// <param name="projectLocation">Path of the project</param>
+        /// <returns>The value of the `OutputType` property</returns>
         public static string LookupOutputTypeFromProjectFile(string projectLocation)
         {
-            var projectFile = FindProjectFileInDirectory(projectLocation);
-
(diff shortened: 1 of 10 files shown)


src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -195,14 +195,67 @@ public static string LookupTargetFrameworkFromProjectFile(string projectLocation
...
-        public static string LookupOutputTypeFromProjectFile(string projectLocation)
+        /// <summary>
+        /// Uses `dotnet msbuild` to determine project properties.
+        /// </summary>
+        /// <param name="projectLocation">Path of the project</param>
+        /// <param name="propertyName">The name of the property</param>
+        /// <returns>The value of the project property</returns>
+        private static string LookupMsbuildProperty(string projectLocation, string propertyName)
         {
-            var projectFile = FindProjectFileInDirectory(projectLocation);
```
normj, line 200: The `dotnet msbuild -getProperty` feature is only available for .NET 8+. We could have users especially in CI systems that are only running with .NET 6. You need to wrap this in a compilation conditional and fallback to the old system when not using .NET 8.

Another option is to always try this approach but if it fails then fallback to the old approach. That would allow us to use this technic inside Visual Studio. But you need to make sure we detect the error correctly.

normj (changes requested):

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
philasmar: I went with the 2nd approach. I will always try to use `dotnet msbuild -getProperty`, but if you are using a .NET SDK prior to NET8, that command will throw the following error: `MSBUILD : error MSB1001: Unknown switch.`. This is thrown when you use: `-getProperty`. In .NET8, if you use that switch with a property type that doesn't exist, the CLI returns empty. It doesn't throw this error.

ashovlin: Possible alternative approach using `Microsoft.Build` and `Micorosft.Build.Locator`.

I had looked at this for Lambda Annotations, but couldn't use that there because source generators have to target .NET Standard and these packages don't. 

I haven't explored all the tradeoffs, but might be nice to not have to manage our own `Process`. There are some implications around Visual Studio compatibility when selecting versions of these dependencies.

```
using Microsoft.Build.Evaluation;
using Microsoft.Build.Locator;

var instance = MSBuildLocator.QueryVisualStudioInstances().OrderByDescending(
       instance => instance.Version).First(); // select which of the installed instances to use

// Register a specific instance of MSBuild
MSBuildLocator.RegisterInstance(instance);

DetectProjectType("path\\to\\project.csproj");

static void DetectProjectType(string pathToCsproj)
{
    var project = new Project();

    var outputType = project.GetProperty("OutputType");

    Console.WriteLine(outputType.EvaluatedValue);
}
```

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -195,14 +195,78 @@ public static string LookupTargetFrameworkFromProjectFile(string projectLocation
...
+            var process = new Process();
+            string outputJson;
+            try
+            {
+                process.StartInfo = new ProcessStartInfo()
+                {
+                    FileName = "dotnet",
+                    Arguments = $"msbuild {projectLocation} -getProperty:{propertyName}",
```
ashovlin, line 213: Possible alternative approach using `Microsoft.Build` and `Micorosft.Build.Locator`.

I had explored this for Lambda Annotations, but couldn't use that there because source generators have to target .NET Standard and these packages don't. 

I haven't explored all the tradeoffs, but might be nice to not have to manage our own `Process` and the .NET 8+ issue Norm mentioned. There are some implications around Visual Studio compatibility when selecting versions of these dependencies.

```
using Microsoft.Build.Evaluation;
using Microsoft.Build.Locator;

var instance = MSBuildLocator.QueryVisualStudioInstances().OrderByDescending(
       instance => instance.Version).First(); // select which of the installed instances to use

// Register a specific instance of MSBuild
MSBuildLocator.RegisterInstance(instance);

// You have to register in a different method than when you start consuming it.
DetectProjectType("path\\to\\project.csproj");

static void DetectProjectType(string pathToCsproj)
{
    var project = new Project(pathToCsproj);

    var outputType = project.GetProperty("OutputType");

    Console.WriteLine(outputType.EvaluatedValue);
}
```

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -195,14 +195,78 @@ public static string LookupTargetFrameworkFromProjectFile(string projectLocation
...
+            {
+                // swallow any exceptions related to `dotnet msbuild`
+                return null;
+            }
 
-            var xdoc = XDocument.Load(projectFile);
+            // If it has completed but unsuccessfully, give up
+            if (process.ExitCode != 0)
+            {
+                if (outputJson.Contains("MSBUILD : error MSB1001: Unknown switch."))
```
normj, line 239: Lets move the fallback logic here and use it anytime the `ExitCode` is non `0`. That will reduce the risk of users seeing any change of behavior. For now I'm fine with just the look up for `OutputType` only using this logic. Once we publish and have and we haven't seen any issues from the change we can switch other msbuild property look ups to this method.

normj (changes requested):

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
normj: This method makes me nervous of something not working write on mac and linux platforms. The locator seems a little flaky for a component automatically picked up in people's CI systems.

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
philasmar: Done

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -195,14 +195,62 @@ public static string LookupTargetFrameworkFromProjectFile(string projectLocation
...
+        /// <summary>
+        /// Retrieve the `OutputType` property of a given project
+        /// </summary>
+        /// <param name="projectLocation">Path of the project</param>
+        /// <returns>The value of the `OutputType` property</returns>
         public static string LookupOutputTypeFromProjectFile(string projectLocation)
         {
-            var projectFile = FindProjectFileInDirectory(projectLocation);
+            var process = new Process();
+            var outputJson = string.Empty;
```
ashovlin, line 206: nit: As long as we're just getting a single property, it should be the value. I'd drop `Json` from the variable name to avoid confusion.
> If you use -getProperty to request a single property, the output is emitted as a string of text. That means you can easily consume the value in scripting scenarios, or within a more complex command line.

test/Amazon.Lambda.Tools.Test/DeployTest.cs
```diff
@@ -186,6 +186,41 @@ public async Task RunDeploymentNativeAot(bool multipleProjects, bool customMount
             }
         }
 
+        [Fact]
+        public async Task NativeAotNet8WebApp()
```
ashovlin, line 190: Could you add a comment with the context on why this was added, mentioning that this test ensures we're detecting the inferred `<OutputType>Exe</OutputType>` correctly?

normj (approved):
Approving assuming you handle @ashovlin comments.

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
philasmar: Done

test/Amazon.Lambda.Tools.Test/DeployTest.cs
philasmar: Done

ashovlin (approved):

PR #309, opened 2024-04-03
Skip NET8 Native AOT test because it's not supported in the pipeline

philasmar:
*Description of changes:*
After some investigation, the NET8 Native AOT test was failing in the pipeline because it could not find the Docker CLI. This is due to the tests running on a Windows container which does not have Docker available on it.
I have skipped the test, similar to what we have done with others tests.

### DIFF
--- a/test/Amazon.Lambda.Tools.Test/DeployTest.cs
+++ b/test/Amazon.Lambda.Tools.Test/DeployTest.cs
@@ -190,7 +190,7 @@ namespace Amazon.Lambda.Tools.Test
         /// This test deploys a .NET8 Native AOT Web App that does not explicitly set the OutputType to Exe.
         /// In .NET8, the web app inherits the OutputType from the SDK.
         /// </summary>
-        [Fact]
+        [Fact(Skip = "This test requires Docker to be available. We are using a Windows container in the release pipeline which does not support Docker.")]
         public async Task NativeAotNet8WebApp()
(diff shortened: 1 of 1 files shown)


test/Amazon.Lambda.Tools.Test/DeployTest.cs
```diff
@@ -190,7 +190,7 @@ public async Task RunDeploymentNativeAot(bool multipleProjects, bool customMount
         /// This test deploys a .NET8 Native AOT Web App that does not explicitly set the OutputType to Exe.
         /// In .NET8, the web app inherits the OutputType from the SDK.
         /// </summary>
-        [Fact]
+        [Fact(Skip = "This test requires Docker to be available. We are using a Windows container in the release pipeline which does not support Docker.")]
```
ashovlin, line 193: Does the same check we use above apply here? I'd rather leave it on-by-default for developers working locally if possible.

https://github.com/aws/aws-extensions-for-dotnet-cli/blob/ea50d25688398071096d7a574eb814a2fedfce3c/test/Amazon.Lambda.Tools.Test/DeployTest.cs#L45-L49

philasmar: Fixed

ashovlin (approved):

normj (approved):

Issue #290, opened 2023-09-07
Extensions CLI publishes beanstalk environments with IMDSv1 enabled


shruti0085:
### Describe the feature

Creating an issue here to track https://github.com/aws/aws-toolkit-visual-studio/issues/359 on the Visual Studio Toolkit.
The Visual Studio Toolkit leverages the AWS Extensions for Dotnet CLI to power some of the publish experiences including legacy beanstalk deploy support, so any changes to behavior will have to originate in this repository.

Beanstalk environments deployed using the CLI creates environment with IMDSv1 enabled instead of IMDSv2 which is now recommended by AWS.

### Use Case

Snippet from the original issue
IMDSv1 has been superseded by IMDSv2 and AWS is suggesting to upgrade to it everywhere, since IMDSv2 is inherently more secure. I noticed that when publishing applications with AWS Toolkit to Elastic Beanstalk, the EC2 instances have the IMDSv1 property activated.

### Proposed Solution

Support creating IMDSv2 enabled beanstalk environments

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Targeted .NET platform

Unavailable

### Environment details (OS name and version, etc.)

Windows

ashishdhingra:
There is a related issue for .NET Deploy tool as well.

ashishdhingra:
[Amazon.ElasticBeanstalk.Tools version 4.4.0](https://www.nuget.org/packages/Amazon.ElasticBeanstalk.Tools/4.4.0) adds support for new boolean command line parameter `--disable-imds-v1` which could be set to `true` to disable IMSDv1.

PR #323, opened 2024-06-20
Added support for new command line parameter --disable-imds-v1 to disable IMDSv1 for Elastic BeanStalk environments.

ashishdhingra:
*Issue #, if available:* #290 (internal issue: DOTNET-7165)

*Description of changes:*
Adds feature to disable `IMDSv1` by default for **new** Elastic BeanStalk environments. Also adds support for new command line parameter `--disable-imds-v1` useful for updating existing environments.

Refer https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environments-cfg-ec2-imds.html#environments-cfg-ec2-imds.namespace for details on how to disable IMDSv1.

**EDIT:** (Per review comment from @normj) We should not disable IMDSv1 by default for new environments since it could break existing users using legacy Visual Studio tooling. The AWS Toolkit for Visual Studio doesn't have checkbox for the new flag and hence users might not know if IMDSv1 is being disabled by default.

**Testing:**
The changes were tested in personal AWS account (via Visual Studio debugging and observing state in AWS EB console) for below scenarios (executed AWS .NET Extensions CLI command):
- Create new EB environment with `--disable-imds-v1` set to `true`.
- Created EB environment with default settings, which could enable IMDSv1.
  - Re-deployed the EB environment with `--disable-imds-v1` set to `true`. This executed update EB environment scenario and disabled IMDSv1.

After this PR is released we also need to create PR to push CLI extensions changes to AWS Toolkit for Visual Studio repository and thereafter their team could decide to add GUI checkbox for the new setting.
___

### DIFF
--- a/src/Amazon.ElasticBeanstalk.Tools/Commands/CommandProperties.cs
+++ b/src/Amazon.ElasticBeanstalk.Tools/Commands/CommandProperties.cs
@@ -19,6 +19,7 @@ namespace Amazon.ElasticBeanstalk.Tools.Commands
         public string IISWebSite { get; set; }
         public bool? WaitForUpdate { get; set; }
         public bool? EnableXRay { get; set; }
+        public bool? DisableIMDSv1 { get; set; }
         public Dictionary<string,string> Tags { get; set; }
         public Dictionary<string, string> AdditionalOptions { get; set; }
 
@@ -92,6 +93,8 @@ namespace Amazon.ElasticBeanstalk.Tools.Commands
                 this.LoadBalancerType = tuple.Item2.StringValue;
             if ((tuple = values.FindCommandOption(EBDefinedCommandOptions.ARGUMENT_ENABLE_STICKY_SESSIONS.Switch)) != null)
                 this.EnableStickySessions = tuple.Item2.BoolValue;
+            if ((tuple = values.FindCommandOption(EBDefinedCommandOptions.ARGUMENT_DISABLE_IMDS_V1.Switch)) != null)
+                this.DisableIMDSv1 = tuple.Item2.BoolValue;
 
             if ((tuple = values.FindCommandOption(EBDefinedCommandOptions.ARGUMENT_PROXY_SERVER.Switch)) != null)
                 this.ProxyServer = tuple.Item2.StringValue;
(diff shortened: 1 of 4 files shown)


src/Amazon.ElasticBeanstalk.Tools/Commands/DeployEnvironmentCommand.cs
```diff
@@ -415,7 +419,14 @@ private async Task<string> CreateEnvironment(string application, string environm
                     Value = loadBalancerType
                 });
             }
-           
+
+            // For new environments, disable IMDSv1 by default (unless explicitly enabled in additional options).
+            createRequest.OptionSettings.Add(new ConfigurationOptionSetting()
```
normj, line 424: Why does this need to be both the here and in `AddAdditionalOptions`? Seems redundant and you had to add code in `AddAdditionalOptions` to handle the fact that it was potentially added by the parent caller.

src/Amazon.ElasticBeanstalk.Tools/Commands/CommandProperties.cs
```diff
@@ -92,6 +93,8 @@ internal void ParseCommandArguments(CommandOptions values)
                 this.LoadBalancerType = tuple.Item2.StringValue;
             if ((tuple = values.FindCommandOption(EBDefinedCommandOptions.ARGUMENT_ENABLE_STICKY_SESSIONS.Switch)) != null)
                 this.EnableStickySessions = tuple.Item2.BoolValue;
+            if ((tuple = values.FindCommandOption(EBDefinedCommandOptions.ARGUMENT_DISABLE_IMDS_V1.Switch)) != null)
```
normj, line 96: The `PersistSettings` method also needs to be updated for the new switch.
https://github.com/aws/aws-extensions-for-dotnet-cli/blob/master/src/Amazon.ElasticBeanstalk.Tools/Commands/CommandProperties.cs#L103

normj (changes requested):

src/Amazon.ElasticBeanstalk.Tools/Commands/DeployEnvironmentCommand.cs
normj: I misunderstood the purpose here which was to force all new deployments to have disable to true. We don't want to do that because that could break users and we don't have a UI element in Visual Studio for users to toggle this.

@ashishdhingra is going to change the logic to remove this section and in the `AddAdditionalOptions` only set the disable if the user explicitly set the field. The VS toolkit can decide if they want to add a checkbox users can see and default the value to `true`.

src/Amazon.ElasticBeanstalk.Tools/Commands/CommandProperties.cs
ashishdhingra: Done in latest commit

src/Amazon.ElasticBeanstalk.Tools/Commands/DeployEnvironmentCommand.cs
ashishdhingra: Removed the default disabling of IMDSv1 since it could break existing customers using legacy Visual Studio tooling. Also simplified code in `AddAdditionalOptions` to not check if existing setting is already added.

normj (approved):

ashovlin (commented):
Changes look fine, but can you add details to the PR description about how you tested it?

We do have some tests asserting against boolean options, such as here: https://github.com/aws/aws-extensions-for-dotnet-cli/blob/a0be75a09dcba3a27548cf12200c71f9d2f0cb75/test/Amazon.ElasticBeanstalk.Tools.Test/DeployTests.cs#L110

ashovlin (approved):

PR #342, opened 2024-11-05
fix: fixed an issue causing lambda package to hang in CI/CD

philasmar:
*Description of changes:*
* Applied a workaround for the hanging `dotnet publish` process which hangs on `Process.WaitForExit()` because of a dotnet tooling bug. There are issues on Microsoft's repos such as this one https://github.com/dotnet/msbuild/issues/2981. The suggested workaround is to add a very large timeout which solves the issue. I have tested it in our pipelines and the tests ran blazingly fast.

### DIFF
--- a/.autover/changes/e52deae1-471a-4119-b874-5ef0ccf690b3.json
+++ b/.autover/changes/e52deae1-471a-4119-b874-5ef0ccf690b3.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.Tools",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Fixed an issue causing 'lambda package' to hang when ran in CI/CD."
+      ]
+    }
+  ]
+}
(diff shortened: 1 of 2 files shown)


96malhar (approved):

normj (changes requested):
Copying a side conversation. I think we need to better understand the hanging before making this change. The latest version of this tool is pulled into a lot of CI builds so we need to be really careful. Also I don't like that the new user experience wouldn't show any output till the dotnet command is complete. For a big project that could take awhile for the build to complete and see the output.

philasmar:
> Copying a side conversation. I think we need to better understand the hanging before making this change. The latest version of this tool is pulled into a lot of CI builds so we need to be really careful. Also I don't like that the new user experience wouldn't show any output till the dotnet command is complete. For a big project that could take awhile for the build to complete and see the output.

I changed the approach to a workaround suggested in https://github.com/dotnet/msbuild/issues/2981.

normj (approved):

ghost (approved):

Kuinox:
I don't know your use case but the danger of this workaround is getting a truncated stdout/stderr.

Issue #211, opened 2022-04-22
Directory.build.props heirarchy is not respected


toddlang:
### Describe the bug

When packaging a lambda using `dotnet lambda package`, if the target project does not include a `<TargetFramework>` node (because it is defined in a Directory.build.props at a higher location) the packaging tool fails saying it does not know what framework to use.

`Missing required parameter: --framework`

### Expected Behavior

The tool will understand that MSBuild can walk directories searching for Directory.build.props files to define these values.

### Current Behavior

The tool fails to walk the folder heirarchy to find the values necessary for the build.

### Reproduction Steps

Create a csproj file for a lambda without &lt;TargetFramework&gt;.
In a directory higher (say, next to your sln file) define a 'Directory.build.props' file and add a <TargetFramework> node there.

This behaviour is all defined here:
https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022

### AWS .NET SDK and/or Package version used

dotnet lambda tools 5.3.0

### Targeted .NET Platform

.Net 6

### Operating System and version

Windows 10

paul-michalik:
In connection with AWS SAM this is even worse. If you set for example your directories for temporary and binary directories, or the language version on global level via `Directory.build.props` then this is completely by AWS SAM and by `dotnet lambda` commands.

chrisoverzero:
This is something I've tackled in #183.

ashovlin:
Small note: we explored using the new `dotnet msbuild -getProperty` that was introduced in 17.8 (8.0.1) to detect the target framework in our Lambda Annotations library. We ended up taking a different approach since we were inside a source generator and can access the TFM that way, but linking the abandoned code here in case it's useful.

Here's the commit for reference: https://github.com/aws/aws-lambda-dotnet/pull/1658/commits/357bf9963cb88081e22ff4cb409eac9a7a15cd6e

normj:
Version 5.12.1 was released today that if the .NET 8 SDK+ is installed the `dotnet msbuild -getProperty` command will be used to evaluate the MSBuild properties like TargetFramework. That will all settings being set in other MSBuild locations like Directory.build.props will be taken into account when determining the framework.

PR #348, opened 2024-11-18
Use MSBuild to find target framework

ghost:
*Issue #, if available:* https://github.com/aws/aws-extensions-for-dotnet-cli/issues/211

*Description of changes:*

When packaging a lambda using dotnet lambda package, if the target project does not include a <TargetFramework> node (because it is defined in a Directory.build.props at a higher location) the packaging tool fails saying it does not know what framework to use.
```
Missing required parameter: --framework
```

### Expected Behavior
The tool will understand that MSBuild can walk directories searching for Directory.build.props files to define these values.

### Current Behavior
The tool fails to walk the folder hierarchy to find the values necessary for the build.

### Changes
1. Creates a generic helper function to get properties for a project using MSbuild and then XML parsing of the csproj as a fallback, https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022
2. Updates find framework function to use above function
3. Updates other existing functions to use the generic function

### Testing
1. Created new lambda function which has the TargetFramework defined in Directory.Build.props instead of the .csproject file. I then added this to the existing unit test case to verify the target framework is read correctly.


src/Amazon.Common.DotNetCli.Tools/Assets/AmazonCommonDotNetCliTools.targets
```diff
@@ -0,0 +1,10 @@
+<Project>
```
normj, line 1: Why do need to do have our own targets file instead of just using the CLI `dotnet msbuild -getProperty:OutputType:TargetFramework,TargetFrameworks` command?

gcbeattyAWS:
```
dotnet msbuild .\testapps\TestFunction\TestFunction.csproj -getProperty:OutputType:TargetFramework,TargetFrameworks
```

was returning 

```
{
  "Properties": {
    "OutputType:TargetFramework": "",
    "TargetFrameworks": ""
  }
}
``` 
for me. even though the csproj file looks like

```
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <AssemblyName>TestFunction</AssemblyName>
    <OutputType>Library</OutputType>
    <PackageId>TestFunction</PackageId>
    <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
    <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
    <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Amazon.Lambda.Serialization.Json" Version="2.0.0" />
  </ItemGroup>
</Project>
```

gcbeattyAWS: looks like it has to be `dotnet msbuild .\testapps\TestFunction\TestFunction.csproj  --getProperty:TargetFramework`. ill play around with this and see if the built in one works

normj: You should be using `TargetFramework` not `OutputType:TargetFramework`. The `OutputType` was used before when we were trying to see if the project was a class library or an executable.

gcbeattyAWS: should be fixed now

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -208,11 +209,72 @@ public static string DeterminePublishLocation(string workingDirectory, string pr
...
+                    CreateNoWindow = true
+                }
+            };
+
+            process.Start();
+            string output = process.StandardOutput.ReadToEnd();
+            string error = process.StandardError.ReadToEnd();
+            process.WaitForExit();
+
+            if (process.ExitCode != 0)
```
normj, line 243: If the process failed we should fallback to the original XML parsing logic. There could be CI systems that are still deploying .NET 6 and only have .NET 6 SDK installed. The `-getoutput` command was added in .NET 8.

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -208,11 +209,72 @@ public static string DeterminePublishLocation(string workingDirectory, string pr
         public static string LookupTargetFrameworkFromProjectFile(string projectLocation)
         {
             var projectFile = FindProjectFileInDirectory(projectLocation);
+            if (string.IsNullOrEmpty(projectFile))
+            {
+                throw new FileNotFoundException("Could not find a project file in the specified directory.");
+            }
 
-            var xdoc = XDocument.Load(projectFile);
+            var arguments = new[]
```
normj, line 217: We look up `TargetFramework`, `OutputType`, and `PublishAot` from the project file in this class. I would like to see this logic generalized so that we can have a method that takes in the project and a `params` of property names and returns back a dictionary of values. I don't really want 3 different code paths starting the `dotnet` process which will end up drifting.

The shared method should have the fallback logic I describe below that if the process fails parse the project xml like it does today so we don't risk any breaking behavior.

testapps/TestFunctionBuildProps/TestFunctionBuildProps/Readme.md
```diff
@@ -0,0 +1,49 @@
+# AWS Lambda Empty Function Project
```
normj, line 1: We don't really need the readme for checked in for the test apps.

normj (changes requested):

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
gcbeattyAWS: ah good point! i will update accordingly.

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
gcbeattyAWS: updated in latest commit

testapps/TestFunctionBuildProps/TestFunctionBuildProps/Readme.md
gcbeattyAWS: updated in latest commit

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -205,72 +206,125 @@ public static string DeterminePublishLocation(string workingDirectory, string pr
...
+                }
+            };
+            try
+            {
                 process.Start();
-                output = process.StandardOutput.ReadToEnd();
-                var hasExited = process.WaitForExit(5000);
+                string output = process.StandardOutput.ReadToEnd();
+                string error = process.StandardError.ReadToEnd();
+                process.WaitForExit();
```
normj, line 245: Should we keep the 5 second timeout or at least some timeout? I'm thinking of @philasmar recent PR that at least added `int.MaxValue` to avoid deadlocks in this API. https://github.com/aws/aws-extensions-for-dotnet-cli/pull/342

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -205,72 +206,125 @@ public static string DeterminePublishLocation(string workingDirectory, string pr
...
+        /// <param name="projectLocation">The location of the project file.</param>
+        /// <returns>The target framework of the project.</returns>
+        public static string LookupTargetFrameworkFromProjectFile(string projectLocation)
+        {
+            var properties = LookupProjectProperties(projectLocation, "TargetFramework", "TargetFrameworks");
+            if (properties.TryGetValue("TargetFramework", out var targetFramework) && !string.IsNullOrEmpty(targetFramework))
+            {
+                return targetFramework;
+            }
+            if (properties.TryGetValue("TargetFrameworks", out var targetFrameworks) && !string.IsNullOrEmpty(targetFrameworks))
```
normj, line 312: The current system would throw an error if `TargetFrameworks` is used. I can see adding support for `TargetFrameworks` but only if resolves to a single value. We shouldn't just pick a framework which could be wrong but not discovered till after deployment. I'm sure if `TargetFrameworks` is used then it is rarely a single value but I have seen it sometimes after refactoring efforts. Say you drop a target but don't get around to changing back to `TargetFramework`.

So short response, if there is more then one framework in `TargetFrameworks` return `null` to stay consistent with previous behavior. In interactive mode that will prompt the user to choose the framework.

normj (changes requested):

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
gcbeattyAWS: sounds good i will update to that logic

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
gcbeattyAWS: sounds good i will update to 5 seconds

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -205,72 +206,129 @@ public static string DeterminePublishLocation(string workingDirectory, string pr
...
+                        if (propertiesElement.TryGetProperty(property, out JsonElement propertyValue))
+                        {
+                            properties[property] = propertyValue.GetString();
+                        }
+                    }
+                }
+                else
                 {
-                    msbuildProcessFailed = true;
+                    Console.WriteLine($"MSBuild process exited with code {process.ExitCode}. Error: {error}");
```
normj, line 263: No `Console.WriteLine` calls are used. This same code is used inside Visual Studio and there is no console there. If you want to log something you need to pass in an instance if `IToolLogger` to handle logging.

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -205,72 +206,129 @@ public static string DeterminePublishLocation(string workingDirectory, string pr
...
+                    // Fallback to XML parsing
+                    properties = LookupProjectPropertiesFromXml(projectFile, propertyNames);
                 }
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-                // swallow any exceptions related to `dotnet msbuild`
-                msbuildProcessFailed = true;
+                Console.WriteLine($"Error executing MSBuild or parsing output: {ex.Message}");
```
normj, line 270: No `Console.WriteLine` calls are used. This same code is used inside Visual Studio and there is no console there. If you want to log something you need to pass in an instance if `IToolLogger` to handle logging.

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -205,72 +206,129 @@ public static string DeterminePublishLocation(string workingDirectory, string pr
...
+                    var element = xdoc.XPathSelectElement($"//PropertyGroup/{propertyName}");
+                    if (element != null && !string.IsNullOrWhiteSpace(element.Value))
+                    {
+                        properties[propertyName] = element.Value;
+                    }
+                }
             }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"Error parsing project file XML: {ex.Message}");
```
normj, line 295: No `Console.WriteLine` calls are used. This same code is used inside Visual Studio and there is no console there. If you want to log something you need to pass in an instance if `IToolLogger` to handle logging.

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -205,72 +206,129 @@ public static string DeterminePublishLocation(string workingDirectory, string pr
...
+                process.WaitForExit(5000);
 
-                // If it hasn't completed in the specified timeout, stop the process and give up
-                if (!hasExited) 
+                if (process.ExitCode == 0)
                 {
-                    process.Kill();
-                    msbuildProcessFailed = true;
-                }
+                    using JsonDocument doc = JsonDocument.Parse(output);
```
normj, line 249: I might be missing something but if propertyNames is a single value isn't the output of the command just the value and not the JSON document?

normj (changes requested):

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
gcbeattyAWS:
```
PS C:\dev\repos\aws-extensions-for-dotnet-cli> dotnet msbuild .\testapps\TestFunction\TestFunction.csproj  --getProperty:TargetFramework,TargetFrameworks
{
  "Properties": {
    "TargetFramework": "net6.0",
    "TargetFrameworks": ""
  }
}
```

is what the commmand returns

.autover/changes/8a5204e4-83ec-45ac-8c75-27512ae9cdf8.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.Tools",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Use MSBuild to find target framework"
```
philasmar, line 7: Add more explanation here for the significance of this for customers. Not sure everyone would know what this implies.

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -205,72 +206,129 @@ public static string DeterminePublishLocation(string workingDirectory, string pr
...
-            var xdoc = XDocument.Load(projectFile);
-
-            var element = xdoc.XPathSelectElement("//PropertyGroup/TargetFramework");
-            return element?.Value;
-        }
-
-        /// <summary>
-        /// Retrieve the `OutputType` property of a given project
+        // <summary>
+        /// Looks up specified properties from a project file.
```
philasmar, line 211: nit: drop "file" since MSBUILD is not looking at the project file (.csproj) alone

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
philasmar: We need to check if the count of propertyNames is 1 or greater since the result varies. If you run:
```
dotnet msbuild --nologo --getProperty:TargetFramework
```
The result will be 
```
net8.0
```
If you run
```
dotnet msbuild --nologo --getProperty:TargetFramework,TargetFrameworks
```
The result will be:
```
{
  "Properties": {
    "TargetFramework": "net8.0",
    "TargetFrameworks": ""
  }
}
```
You need to handle both cases

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
philasmar: I'm not too sure we should be logging anything visible to the user here. We are exposing an implementation detail that would be more confusing if anything.

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
philasmar: Same comment about logging

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
philasmar: Same comment about logging

philasmar (changes requested):

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
gcbeattyAWS: ah yes you are right. i will update that!

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
gcbeattyAWS: i just removed the logging all together

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
gcbeattyAWS: updated to handle case for when only one property is provided. Thanks for catching this!

normj (approved):
One minor comment to address an existing issue but I don't need to review the PR again assuming you address the comment.

src/Amazon.Common.DotNetCli.Tools/Utilities.cs
```diff
@@ -288,43 +353,32 @@ public static bool LookPublishAotFlag(string projectLocation, string msBuildPara
...
-            var xdoc = XDocument.Load(projectFile);
-
-            var element = xdoc.XPathSelectElement("//PropertyGroup/SelfContained");
-
-            if (bool.TryParse(element?.Value, out _))
+            var properties = LookupProjectProperties(projectLocation, "SelfContained");
+            if (properties.TryGetValue("SelfContained", out var selfContained))
             {
-                return true;
+                return bool.TryParse(selfContained, out _);
```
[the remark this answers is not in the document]
gcbeattyAWS: I am assuming the reason we are allowed to update this without worrying about backward compatibility is because if user's wanted to set the value as `false` previously, then they would have had to not include this property in project file itself. So since we are updating the logic to read the value now, it will not break anything for those customers, since they would not have had the property in the file anyway.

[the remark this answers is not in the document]
normj: I can't think of a reason users would rely on having `<SelfContainted>false</SelfContainted>` and are more likely to cause confusion. I'm also guess most users are relying on the commandline switch `--self-contained` instead of the msbuild property. So I feel safe make this change.

normj (approved):

philasmar (approved):

PR #349, opened 2024-11-19
Sync Master to Dev

normj:
*Description of changes:*
Sync the dev branch after releasing the SnapStart work for Amazon.Lambda.Tools

### DIFF
--- a/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
+++ b/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
@@ -60,6 +60,8 @@ namespace Amazon.Lambda.Tools.Commands
             LambdaDefinedCommandOptions.ARGUMENT_LOG_APPLICATION_LEVEL,
             LambdaDefinedCommandOptions.ARGUMENT_LOG_SYSTEM_LEVEL,
             LambdaDefinedCommandOptions.ARGUMENT_LOG_GROUP,
+
+            LambdaDefinedCommandOptions.ARGUMENT_SNAP_START_APPLY_ON
         });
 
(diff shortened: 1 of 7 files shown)


src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -703,6 +709,20 @@ private UpdateFunctionConfigurationRequest CreateConfigurationRequestIfDifferent
...
 
+            var snapStartApplyOn = this.GetStringValueOrDefault(this.SnapStartApplyOn, LambdaDefinedCommandOptions.ARGUMENT_SNAP_START_APPLY_ON, false);
+            if (!string.IsNullOrEmpty(snapStartApplyOn))
+            {
+                if (null == request.SnapStart)
+                    request.SnapStart = new SnapStart();
+
+                if (!string.Equals(existingConfiguration?.SnapStart?.ApplyOn?.Value, snapStartApplyOn, StringComparison.Ordinal))
+                {
+                    request.SnapStart.ApplyOn = snapStartApplyOn;
```
boblodgett, line 720: Can request.SnapStart be null here based on the if check?

boblodgett (approved):

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
normj: `existingConfiguration.SnapStart` is right. We are checking to see if the specified by the user value `snapStartApplyOn` is different what is current set on the function ``existingConfiguration.SnapStart`. If they are then set the new value on the patch update `request` object.

Issue #358, opened 2024-12-24
The --msbuild-parameters flag is performing incorrectly when spawned by child_process


Levelleor:
### Describe the bug

I am executing the following command via a JS Github Action:
```
dotnet lambda package --disable-interactive True --project-location /sources/dotnet/lambda/src --output-package publish/output.zip --function-runtime dotnet8 --function-architecture arm64 --msbuild-parameters "--no-restore --no-build"
```

I need the ability to pass multiple arguments via msbuild-parameters flag and thus I've encapsulated them into double quotes as requested by the help manual. Both --no-restore and --no-build here are appended as a single command-line argument and are not parsed by dotnet correctly, instead being set on MSBuild call and failing the execution.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

I expect this to be called:
```
dotnet publish "/sources/dotnet/lambda/src" --output "/sources/dotnet/lambda/src/bin/Release/net8.0/publish" --configuration "Release" --framework "net8.0" --no-restore --no-build /p:GenerateRuntimeConfigurationFiles=true --runtime linux-arm64 --self-contained False 
```

### Current Behavior

I get this result instead:
```
dotnet publish "/sources/dotnet/lambda/src" --output "/sources/dotnet/lambda/src/bin/Release/net8.0/publish" --configuration "Release" --framework "net8.0" "--no-restore --no-build" /p:GenerateRuntimeConfigurationFiles=true --runtime linux-arm64 --self-contained False 
```
.NET isn't parsing --no-restore and --no-build because it didn't recognize them and thus it passed these down to msbuild it seems like.

```
... publish: MSBUILD : error MSB1001: Unknown switch.
[...]

PR #360, opened 2025-01-08
Added logic to handle extraneous double-quotes around value for --msbuild-parameters that could be passed in .NET argument for certain execution environments.

ashishdhingra:
*Issue #, if available:* #358

*Description of changes:*
Added logic to handle extraneous double-quotes around value for `--msbuild-parameters` that could be passed in .NET argument for certain execution environments.

Customer is invoking `dotnet lambda` tooling via JavaScript running using `exec.exec()` from `actions/github-script@v6` package in a GitHub action. For this environment, somehow value for `--msbuild-parameters` enclosed in double quotes has leading and trailing double quotes. Did below testing as well:
- Removed use of double quotes all together, it only takes first value for `--msbuild-parameters`, ignoring other values if any.
- Tried enclosing value in single quotes `'`, it tries to parse only first value beginning with single quote (e.g. `'----no-restore`).

**NOTE(s):**
- [CommandOptions](https://github.com/aws/aws-extensions-for-dotnet-cli/blob/fa1385e31a89ace8f8bf819ac5892953464e2592/src/Amazon.Common.DotNetCli.Tools/Options/CommandOptions.cs#L12) also has separate property named `MSBuildParameters` used for storing parameters that start with `/p:` based on logic [here](https://github.com/aws/aws-extensions-for-dotnet-cli/blob/fa1385e31a89ace8f8bf819ac5892953464e2592/src/Amazon.Common.DotNetCli.Tools/Options/CommandLineParser.cs#L70).
  - These MS Build parameters are appended to existing values that might have been supplied using `--msbuild-parameters` switch (e.g., refer code [here](https://github.com/aws/aws-extensions-for-dotnet-cli/blob/fa1385e31a89ace8f8bf819ac5892953464e2592/src/Amazon.Lambda.Tools/Commands/PackageCommand.cs#L112)).
  - Hence, we would need to process trimming of enclosing double quotes `"` before we append `CommandOptions.MSBuildParameters`.
  - Values for consolidated MSBuild parameters are also used at different places while determining project information, e.g. TargetFramework using [Utilitites.LookupTargetFrameworkFromProjectFile()](https://github.com/aws/aws-extensions-for-dotnet-cli/blob/fa1385e31a89ace8f8bf819ac5892953464e2592/src/Amazon.Common.DotNetCli.Tools/Utilities.cs#L319), where it uses [Utilities.LookupProjectProperties()](https://github.com/aws/aws-extensions-for-dotnet-cli/blob/fa1385e31a89ace8f8bf819ac5892953464e2592/src/Amazon.Common.DotNetCli.Tools/Utilities.cs#L217) and invokes `msbuild` command in a separate `Process`. This command would always fail if we don't handle this issue (assuming compatible parameter for `msbuild` command is used. We do have a fallback mechanism to inspect `.csproj` project file though.
- We **should only** handle enclosing double-quote scenario only if `--msbuild-parameters` value start and end with a double-quote. 
- `--msbuild-parameters` is currently only supported by `Amazon.Lambda.Tools`.
  - Based on discussion with @normj, we could only version bump `Amazon.Lambda.Tools`.
___

### DIFF
--- a/test/Amazon.Lambda.Tools.Test/OptionParseTests.cs
+++ b/test/Amazon.Lambda.Tools.Test/OptionParseTests.cs
@@ -81,5 +81,39 @@ namespace Amazon.Lambda.Tools.Test
             Assert.NotNull(param);
             Assert.Equal("us-west-2", param.Item2.StringValue);
         }
+
+        [Fact]
+        public void ParseMSBuildSwitchDoubleQuotedValue()
+        {
+            var values = CommandLineParser.ParseArguments(DeployFunctionCommand.DeployCommandOptions,
+                new[] { "myfunc", "--region", "us-west-2", "--msbuild-parameters", "\"--no-restore --no-build\"" });
+
+            Assert.Equal("myfunc", values.Arguments[0]);
+
+            var msbuildparametersParam = values.FindCommandOption("--msbuild-parameters");
+            Assert.NotNull(msbuildparametersParam);
+            Assert.Equal("--no-restore --no-build", msbuildparametersParam.Item2.StringValue);
+
+            var param = values.FindCommandOption("--region");
+            Assert.NotNull(param);
+            Assert.Equal("us-west-2", param.Item2.StringValue);
+        }
+
+        [Fact]
+        public void ParseMSBuildSwitchNotDoubleQuotedValue()
+        {
+            var values = CommandLineParser.ParseArguments(DeployFunctionCommand.DeployCommandOptions,
+                new[] { "myfunc", "--region", "us-west-2", "--msbuild-parameters", "--no-restore --no-build" });
+
+            Assert.Equal("myfunc", values.Arguments[0]);
+
+            var msbuildparametersParam = values.FindCommandOption("--msbuild-parameters");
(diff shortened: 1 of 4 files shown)


test/Amazon.Lambda.Tools.Test/OptionParseTests.cs
```diff
@@ -81,5 +81,22 @@ public void ParseMSBuildParameters()
             Assert.NotNull(param);
             Assert.Equal("us-west-2", param.Item2.StringValue);
         }
+
+        [Fact]
+        public void ParseMSBuildSwitchDoubleQuotedValue()
```
gcbeattyAWS, line 86: can you add a test case for when -msbuild-parameters does not have double quotes

ashishdhingra: Added. I also added additional test in `test/Amazon.Lambda.Tools.Test/CommandLineParserTest.cs` that has scenario where both `--msbuild-parameters` enclosed in double quote and `/p:` MSBuild parameters are present.

normj (approved):

gcbeattyAWS (approved):

Issue #285, opened 2023-07-14
dotnet eb deploy-environment not setting the Procfile web startup properly


ddivita:
### Describe the bug

I will preface this by saying this could be a GitLab pipeline issue, however, I have also seen this happen on a local development machine.

Command:
```
dotnet eb deploy-environment --application $EB_APPLICATION_NAME --environment $EB_ENVIRONMENT_NAME --version-label $CI_COMMIT_SHA-$CI_JOB_STARTED_AT --self-contained true --region $AWS_DEFAULT_REGION
```
GitLab script:
```
api_deploy_job:
  stage: deploy
  image: mcr.microsoft.com/dotnet/sdk:7.0
  only:
    - develop
  script:
    - echo $CI_COMMIT_SHA
    - echo $AWS_DEFAULT_REGION 
    - apt-get update && apt-get install -y unzip
    - apt-get install -y zip   
    - apt-get update
    - apt-get install -y awscli
    - dotnet --version
    - aws --version
    - dotnet tool install -g Amazon.ElasticBeanstalk.Tools
    - export PATH="$PATH:/root/.dotnet/tools"
    - cd HP.Api         
    - dotnet eb deploy-environment --application $EB_APPLICATION_NAME --environment $EB_ENVIRONMENT_NAME --version-label $CI_COMMIT_SHA-$CI_JOB_STARTED_AT --self-contained true --region $AWS_DEFAULT_REGION
```

The script initiates the dotnet publish command:
```
dotnet publish "/builds/hp-services/HP.Api" --output "/builds/hp-services/HP.Api/bin/Release/net7.0/publish" --configuration "Release" --framework "net7.0"  --runtime linux-x64 --self-contained true
```

After the `publish` compiles the project and creates the files, the `Procfile` is auto-generated. The issue is the `Procfile` is not setting the proper startup `web` assembly.
![image](https://github.com/aws/aws-extensions-for-dotnet-cli/assets/3431250/720bf190-a697-4607-bc3b-db37d912832d)

### Expected Behavior

Running the same command on a local environment produces this Procfile:
![image](https://github.com/aws/aws-extensions-for-dotnet-cli/assets/3431250/ef63aca9-b8c2-4d0f-8930-a644a56dbb3f)

### Current Behavior

The auto-generated `Procfile` is pointing to the wrong assembly. It's not always the same assembly either, but it never picks the API assembly. Even when the `Procfile` is created manually, it still does not deploy properly.

[...]

PR #363, opened 2025-02-19
Fixed an ElasticBeanstalk deployment issue for Linux platform where Procfile was sometimes being generated with incorrect entrypoint when multiple runtimeconfig.json files were present.

ashishdhingra:
*Issue #, if available:* #285

*Description of changes:*
Fixed an ElasticBeanstalk deployment issue for Linux platform where `Procfile` was sometimes being generated with incorrect entrypoint when multiple `runtimeconfig.json` files were present.

The existing logic to determine the project file `.csproj` uses [Utilitites.FindProjectFileInDirectory()](https://github.com/aws/aws-extensions-for-dotnet-cli/blob/b2735be679eeea008d709fc7a57d2fabde66085a/src/Amazon.Common.DotNetCli.Tools/Utilities.cs#L389). It has explicit check [here](https://github.com/aws/aws-extensions-for-dotnet-cli/blob/b2735be679eeea008d709fc7a57d2fabde66085a/src/Amazon.Common.DotNetCli.Tools/Utilities.cs#L397) to ensure that we get only **one** project file; else it returns `null`. So in case there are multiple project files, it would probably throw `NullReferenceException` in subsequent logic. We are using this assumption that we only have one project file for deployment at root of working directory, to get the assembly name (via `msbuild`).

Also note that we also invoke [BaseCommand.EnsureInProjectDirectory()](https://github.com/aws/aws-extensions-for-dotnet-cli/blob/b2735be679eeea008d709fc7a57d2fabde66085a/src/Amazon.Common.DotNetCli.Tools/Commands/BaseCommand.cs#L890). This throws generic exception `$"No .NET project found in directory {projectLocation} to build."` in case where more than one project file if found in working directory, **which might be incorrect or misleading**.

**NOTE:**
Refer [Using a Procfile to configure your .NET Core on Linux Elastic Beanstalk environment](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/dotnet-linux-procfile.html) where it shows below example to configure 2 web applications.
```
web: dotnet ./dotnet-core-app1/dotnetcoreapp1.dll
web2: dotnet ./dotnet-core-app2/dotnetcoreapp2.dll
```
The PR supports one web application as of now to conform to existing logic.
CC @normj 
___

### DIFF
--- a/src/Amazon.ElasticBeanstalk.Tools/EBUtilities.cs
+++ b/src/Amazon.ElasticBeanstalk.Tools/EBUtilities.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using Amazon.Common.DotNetCli.Tools;
+using Amazon.Common.DotNetCli.Tools.Options;
 using Amazon.ElasticBeanstalk.Model;
 using Amazon.ElasticBeanstalk.Tools.Commands;
 using ThirdParty.Json.LitJson;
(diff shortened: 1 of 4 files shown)


src/Amazon.ElasticBeanstalk.Tools/EBUtilities.cs
```diff
@@ -114,20 +115,21 @@ public static void SetupPackageForLinux(IToolLogger logger, EBBaseCommand comman
...
 
-            if(File.Exists(procfilePath))
+            if (File.Exists(procfilePath))
             {
                 logger?.WriteLine("Found existing Procfile file found and using that for deployment");
                 return;
             }
 
             logger?.WriteLine("Writing Procfile for deployment bundle");
+            var projectLocation = Utilities.DetermineProjectLocation(command.WorkingDirectory,
```
normj, line 125: The caller of this method has already used `DetermineProjectLocation` to determine the location. Update this method signature to pass in the project location so we don't need to detect the location again. Since this is a tool not a library we can change the public signature.

normj (changes requested):

normj (approved):

PR #447, opened 2026-06-30
Auto-attach durable execution IAM policy in deploy-function

GarrettBeatty:
## Description

A durable Lambda function's execution role must carry the AWS-managed `AWSLambdaBasicDurableExecutionRolePolicy` so the function can call the durable-execution checkpoint APIs. The `--durable-execution-timeout` / `--durable-retention-period` switches (added previously) set `DurableConfig` on the function, but nothing ensured the role had the required policy — a function deployed with `deploy-function` and a tool-created role would fail at runtime because it couldn't checkpoint.

This PR makes `deploy-function` / `update-function-config` ensure the durable IAM policy is in place whenever the deploy is durable (i.e. either `--durable-*` value resolves).

## Changes

- **`LambdaConstants`**: add `AWS_LAMBDA_BASIC_DURABLE_EXECUTION_MANAGED_POLICY` and register it in `KNOWN_MANAGED_POLICY_DESCRIPTIONS` so it shows up in the interactive role-creation picker.
- **`RoleHelper`**: add idempotent `EnsureManagedPolicyAttachedAsync` / `IsManagedPolicyAttachedAsync` plus a `GetRoleNameFromArnOrName` helper.
- **`DeployFunctionCommand` (create path)**: factor durable config resolution into `TryResolveDurableConfig`; after create, attach the durable policy when the tool created the role, and warn (without mutating) when the role was user-supplied.
- **`UpdateFunctionConfigCommand` (update path)**: warn-only when a durable function's existing role is missing the policy (the tool never creates a role on update).

## Behavior decisions

- **Trigger**: presence of either `--durable-*` value (no new flag) — matches the existing wiring.
- **User-supplied role**: warn only, never mutate IAM the tool doesn't own. IAM read/attach errors are downgraded to warnings so a deploy isn't failed over a missing `iam:AttachRolePolicy` permission.

### Unit tests (fully mocked, no AWS)
- `RoleHelperDurableTests` — name extraction, is-attached true/false, attach-when-missing, idempotent-when-present (7 tests).
- `DurableConfigTest` — `DurableConfig` mapped onto create + update requests; non-durable deploy attaches nothing; user-supplied role missing the policy warns without attaching (4 tests).

All 11 new tests pass; existing `OptionParseTests` still green.

### End-to-end (real AWS, us-east-1, managed `dotnet10` runtime)

Deployed a real durable function (executable model, `DurableFunction.WrapAsync` + a single `StepAsync`) with the locally-built tool and exercised every path:

**Confirmed the problem exists (old behavior).** A durable function deployed on a role with only `AWSLambdaBasicExecutionRole` (what the tool produced before this change) fails at the first checkpoint:

```
Status: FAILED
Error: Failed to checkpoint operations ... User: .../<role> is not authorized to perform:
lambda:CheckpointDurableExecution ... because no identity-based policy allows the
lambda:CheckpointDurableExecution action
```

So before this fix the function deploys fine but fails at runtime, with no warning.

**Attach path (interactive role creation).** Created a fresh role through the prompt and deliberately chose "No policy" — the tool still auto-attached the durable policy:

```
Attached durable execution managed policy arn:aws:iam::aws:policy/service-role/AWSLambdaBasicDurableExecutionRolePolicy to role arn:aws:iam::<acct>:role/<new-role>
```

`list-attached-role-policies` showed the policy present, and the durable execution then ran to **SUCCEEDED**. The new known-policy entry also appeared in the interactive picker.

**Warn path (user-supplied role, `--disable-interactive`).** On both create and update the tool emitted the warning and did **not** mutate the user's role:

```
Warning: durable function role arn:aws:iam::<acct>:role/<role> is missing the
arn:aws:iam::aws:policy/service-role/AWSLambdaBasicDurableExecutionRolePolicy managed policy.
Durable execution checkpointing will fail until this policy is attached to the role.
```

`DurableConfig` (e.g. `ExecutionTimeout=300`) was correctly set on the function in all cases.

**Positive control.** Attaching the durable policy to the previously-failing role and re-invoking → **SUCCEEDED** (`Error: null`). (Note: a cold start + ~60–90s for managed-policy propagation into the execution role's session was needed before the new permission took effect.)

All test functions and IAM roles created during the run were torn down afterward.

> Draft: opening for review of the approach before finalizing.

### DIFF
--- a/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
+++ b/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
@@ -496,7 +496,7 @@ namespace Amazon.Lambda.Tools.Commands
         /// A durable function's execution role must carry the AWSLambdaBasicDurableExecutionRolePolicy managed policy
         /// so the function can call the durable-execution checkpoint APIs. When the tool created the role it attaches
         /// the policy automatically. For a role the user supplied (or an existing function's role on update) the tool
-        /// does not mutate IAM it does not own; it only warns when the policy is missing.
+        /// does not mutate IAM it does not own; it only emits an informational note when the policy is not attached.
         /// </summary>
         /// <param name="roleArn">The resolved role ARN (or name) the function uses.</param>
         /// <param name="toolCreatedRole">True when this deployment created the role, in which case the policy is attached.</param>
@@ -517,16 +517,17 @@ namespace Amazon.Lambda.Tools.Commands
                 }
                 else if (!await RoleHelper.IsManagedPolicyAttachedAsync(this.IAMClient, roleArn, policyArn))
                 {
-                    this.Logger.WriteLine($"Warning: durable function role {roleArn} is missing the {policyArn} managed policy. " +
-                        "Durable execution checkpointing will fail until this policy is attached to the role.");
+                    this.Logger.WriteLine($"Using existing IAM role {roleArn} for Lambda function. Ensure this role has the " +
+                        $"required permissions for Durable Functions. The {policyArn} managed policy can be attached to grant permissions.");
                 }
             }
             catch (Exception e)
             {
-                // Don't fail the deployment over an IAM read/attach issue (e.g. the caller lacks iam:AttachRolePolicy
-                // or iam:ListAttachedRolePolicies on a user-managed role). Surface a warning so the user can act.
-                this.Logger.WriteLine($"Warning: unable to verify or attach the durable execution managed policy {policyArn} " +
-                    $"on role {roleArn}: {e.Message}. Ensure the policy is attached for durable execution to work.");
+                // The deployment role may have no IAM permissions (e.g. the caller lacks iam:AttachRolePolicy or
+                // iam:ListAttachedRolePolicies), which can be a best practice for locking down permissions. Don't fail
+                // the deployment over it; surface the same informational message so the user can act if needed.
+                this.Logger.WriteLine($"Using existing IAM role {roleArn} for Lambda function ({e.Message}). Ensure this role " +
+                    $"has the required permissions for Durable Functions. The {policyArn} managed policy can be attached to grant permissions.");
             }
         }
 
--- a/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
+++ b/src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
@@ -310,6 +310,16 @@ namespace Amazon.Lambda.Tools.Commands
                 }
             }
 
+            // When updating a function with durable configuration, the function's execution role must carry the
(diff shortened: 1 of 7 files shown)


src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -458,6 +468,68 @@ private UpdateFunctionConfigurationRequest CreateRevertConfigurationRequest(GetF
...
+            {
+                if (toolCreatedRole)
+                {
+                    var attached = await RoleHelper.EnsureManagedPolicyAttachedAsync(this.IAMClient, roleArn, policyArn);
+                    if (attached)
+                        this.Logger.WriteLine($"Attached durable execution managed policy {policyArn} to role {roleArn}");
+                }
+                else if (!await RoleHelper.IsManagedPolicyAttachedAsync(this.IAMClient, roleArn, policyArn))
+                {
+                    this.Logger.WriteLine($"Warning: durable function role {roleArn} is missing the {policyArn} managed policy. " +
```
GarrettBeatty: technically the user might have some custom policy that has everything durable execution policy has and we wont catch this. but this is just a best effort to help the user

normj: As you said the function's role might also cover the permission requirement in other ways. I suggest reworking the wording to be less sounding like they are in error and more info based. Something like

```
Using existing IAM role {roleArn} for Lambda function. Ensure this role has the required permissions for Durable Functions. Lambda The {policyArn} managed policy can be attached to grant permissions.
```

src/Amazon.Lambda.Tools/Commands/UpdateFunctionConfigCommand.cs
```diff
@@ -458,6 +468,68 @@ private UpdateFunctionConfigurationRequest CreateRevertConfigurationRequest(GetF
...
+                        "Durable execution checkpointing will fail until this policy is attached to the role.");
+                }
+            }
+            catch (Exception e)
+            {
+                // Don't fail the deployment over an IAM read/attach issue (e.g. the caller lacks iam:AttachRolePolicy
+                // or iam:ListAttachedRolePolicies on a user-managed role). Surface a warning so the user can act.
+                this.Logger.WriteLine($"Warning: unable to verify or attach the durable execution managed policy {policyArn} " +
```
normj: I would reuse the same comment above. It is possible their deployment role has no IAM permissions which can be a best practice for locking down permissions and we don't want them to feel like something is wrong.

GarrettBeatty: https://github.com/aws/aws-extensions-for-dotnet-cli/pull/447/commits/3712b36efa40ab635bea209f7e8b8944f9267843 updated wording here

normj (approved):

philasmar (approved):

PR #451, opened 2026-07-10
Remove hidden zip utility dependency from deployment bundling

GarrettBeatty:
## Problem

`zip` is a hidden, undocumented prerequisite. On a clean Linux environment (e.g. the bare .NET SDK container images, which do not ship `zip`), `dotnet lambda deploy-serverless` — and also `deploy-function` / `package` — fail with:

```
Failed to find the "zip" utility program in path. This program is required to maintain Linux file permissions in the zip archive.
```

## Root cause

Deployment bundling needs the archived files (notably the custom runtime `bootstrap`) to carry Unix execute permissions so Lambda can run them. Those permissions live in each ZIP entry's central-directory *external attributes*, and Linux only honors them when the entry's *"version made by"* host byte is Unix (3).

Historically the tools worked around .NET's inability to set the host byte by shelling out to external utilities:
- **Linux/macOS** → the native `zip` program (the missing prerequisite).
- **Windows** → a vendored Go binary, `build-lambda-zip.exe`.

## Fix

Introduce `ManagedZipArchive`, which creates the archive with `System.IO.Compression` and then rewrites each central-directory record so the host system is reported as Unix and the external attributes carry a `0777` file mode — reproducing exactly what `build-lambda-zip` did (every entry marked executable). This removes the external-process dependency on **all** platforms.

Changes:
- Add `ManagedZipArchive` (in `Amazon.Common.DotNetCli.Tools`).
- Route `Utilities.ZipDirectory` (serverless / Elastic Beanstalk / ECS) and `LambdaPackager.BundleDirectory`/`BundleFiles` (Lambda) through it. `bootstrap` line-ending normalization and runtime-flattening logic are preserved.
- Delete the vendored `build-lambda-zip.exe`, its Go source (`build-lambda-zip.go`, `go.mod`, `go.sum`), the `Content` include and `build-lambda-zip` MSBuild target, and the `zip` entry in the CLI wrapper's known locations.
- The central directory is located via the End Of Central Directory record (not by scanning for the signature bytes) so file content that happens to contain `PK\x01\x02` cannot confuse the patch logic.

## Testing

- New unit tests (`ManagedZipArchiveTests`): content round-trips, every entry is marked Unix host + `0777`, and a file whose bytes contain the central-directory signature is handled correctly.
- Existing suites pass against a real AWS account (Isengard):
  - `LambdaSingleFilePackageTests` (26) — real publish + zip end-to-end.
  - `ArmTests` (4) — includes `TestServerlessPackage` (the `Utilities.ZipDirectory` path) and a deploy.
  - `DeployTest.RunDeployAndInvokeWithIntegerPayloadCommand` — deploy **and invoke**, proving the extracted archive runs.
  - `DeployTest.RunYamlServerlessDeployCommand` + `RunDeployCommand` — the full `deploy-serverless` path.
  - `Amazon.ElasticBeanstalk.Tools.Test.CreatePackageTests` (3) — EB bundling round-trip.

An autover change file is included (patch bump for all three tools); CHANGELOG is generated by autover at release time.


src/Amazon.Common.DotNetCli.Tools/ManagedZipArchive.cs
```diff
@@ -0,0 +1,170 @@
...
+    /// limitation by shelling out to the native "zip" program on Linux/macOS and to a vendored Go
+    /// utility (build-lambda-zip.exe) on Windows.
+    ///
+    /// This class removes both of those external dependencies. It writes the archive with
+    /// System.IO.Compression and then rewrites each central directory record so that the host system is
+    /// reported as Unix and the external attributes carry a 0777 (rwxrwxrwx) file mode. This mirrors the
+    /// behavior of the old build-lambda-zip utility, which unconditionally marked every entry as an
+    /// executable Unix file, so the resulting archives behave identically on Lambda.
+    /// </summary>
+    public static class ManagedZipArchive
```
GarrettBeatty, line 29: claude generated all of this.  i asked it how it knew the constants were correct. heres what it did


" How I confirmed the choices were right

  Rather than trust the spec blindly, during the investigation I wrote a throwaway program, produced a real archive, and dumped the central-directory bytes
  with Python's struct, checking that version_made_by >> 8 == 3 and external_attr >> 16 == 0o100777. Then I round-tripped it through Python's zipfile
  (testzip() returned clean) to confirm the patched bytes still form a valid archive. That's why the offsets are known-correct and not just copied from a
  reference.

GarrettBeatty:
marking as draft until norm is back

